Issue #8, opened 2017-07-02
GCode generation


jeffeb3:
From a set of properties, like {speed, coordinates} the gcode generator should generate the text needed to run the machine. Ideally, this would be available as a download. Anything that will make this slicker, such as a pre or post script should be pushed into the next issue, just so we can get this going faster.

PR #10, opened 2017-07-02
"functioning" gcode generator. The Generate GCode button will read in…

jeffeb3:
… whatever vertices are in the App.this.state.vertices and produce a simple gcode file as a download.

Currently, this is set up to just have a list of 0,0 1,1, and that sorta works.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,8 +14,33 @@ import {
     Panel,
 } from 'react-bootstrap'
 import TurtleCanvas from './TurtleCanvas';
+import GCodeGenerator from './GCode';
+
+// I'm trying to define a simple struct that we can use everywhere we need vertices. I don't see a
+// problem letting this bloat a little.
+//
+// Currently, I'm using this as input to the GCodeGenerator for the locations of gcode.
+//
+function vertex (x, y, speed=0) {
+  return {
+    x: x,
+    y: y,
+    f: speed
+  }
+}
 
 class App extends Component {
+
+  constructor(props) {
+    super(props)
+    this.state = {
+      vertices: [
+        vertex(0.0, 0.0),
+        vertex(1.0, 1.0),
+      ],
+    };
+  }
+
   render() {
 
     function saySquare() {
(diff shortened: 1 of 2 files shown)


jeffeb3:
How do I connect this to issue #8 ?

jeffeb3:
@jamesd256, if you could just take a look and point out where I might be doing terrible javascript/old syntax/unsafe things, and what I would do to fix these things, that would help me and the code get better.

src/App.js
```diff
@@ -1,16 +1,38 @@
...
+    f: speed
+  }
+}
 
 class App extends Component {
+
+  constructor(props) {
+    super(props)
+    this.state = {
+      vertices: [
```
jeffeb3: This list will be empty soon.

src/App.js
```diff
@@ -32,7 +54,9 @@ class App extends Component {
             <TurtleCanvas width={300} height={300} rotation={45}/>
           </div>
 
-          <Button id="gcode" bsStyle="primary">Generate GCode</Button>
+          <div id="output">
+            <GCodeGenerator vertices={this.state.vertices}/>
```
jeffeb3: I want to pass `this.state.vertices` into Shapes and Transforms, or maybe I'll need to just send it into Transforms and have a separate list of vertices come from Shapes and go to Transforms. But they will be responsible for generating the vertices.

I also want to pass the vertices to the canvas and that's how we'll get a live preview.

If/When we have other modes for generating vertices, like a spirograph, or turtle output, they would also create the vertices. They would end up in the canvas and ultimately in the gcode.

src/GCode.js
```diff
@@ -0,0 +1,57 @@
...
+  return command + '\n'
+}
+
+// A class that will encapsulate all the gcode generation. This is currently just a button that
+// converts vertices and a speed into a list of positions. There is a lot more than could exists
+// here.
+class GCodeGenerator extends Component {
+
+  generateGCode() {
+    var content = ""
```
jeffeb3: You can see right here, we are building the gcode document. We could add in the precode and the postcode here.

src/GCode.js
```diff
@@ -0,0 +1,57 @@
...
+    for (var i=0; i<this.props.vertices.length; i++) {
+      content += gcode(this.props.vertices[i]);
+    }
+    download('sandify.gcode', content)
+  }
+
+  render() {
+    return (
+      <Button id="gcode" bsStyle="primary" onClick={this.generateGCode.bind(this)}>Generate GCode</Button>
```
jeffeb3: Currently just a button, but it could be an area for precode, postcode, speeds, formats (like line numbers or base filenames).

Issue #17, opened 2017-07-05
Machine Limits


jeffeb3:
Recompute vertices so that the shape is as preserved as possible when the vertices exit the prescribed machine size.

jeffeb3:
Works great, in the scenarios I've used.

jeffeb3:
I broke it... It works well for big squares, but not for stars, anymore.

GhostDaddio:
I've been working on the machine limits and I think I've got a mostly brute force method.  It mostly works, but when I tried to push it, I was denied access.   How do I do that?  The LimitEnforcer.js file I tried to push is attached.
[LimitEnforcer.zip](https://github.com/jeffeb3/sandify/files/1149996/LimitEnforcer.zip)

jeffeb3:
Thanks. I've been cooking up some progress offline about this, but I appreciate some help. You can't push to this repo (I think you can't push to any branches either). I will look at the zip, but in the future, you'll need to learn how to make a Pull Request. If you're not on the project, then you'll need to fork the project first:

https://help.github.com/articles/creating-a-pull-request-from-a-fork/

BTW, are you someone we know on the forums with a different name?

jeffeb3:
![2017-07-14--1500094046_1920x1050_scrot](https://user-images.githubusercontent.com/1812882/28236589-82c741a0-68e6-11e7-9c14-3cd48aa2d98f.png)

Not a big fan of comments, eh? I think I am one too many beers away from understanding this without some direction, or variable names. I did run it against my first diabolical case (above).

jeffeb3:
I'll push my working branch so you can see what I've been doing. I am segmenting the problem into which of 9 boxes the vertex and the previous vertex lie, because the hard cases are when the vertices go out from one end and enter from another. I think I'm pretty close, I just need to concentrate for a moment on it.

jeffeb3:
I pushed my stuff to limits_2. I usually do a little cleanup when I'm done, and I'm not done yet. I think I only have trouble with the give up section, IIRC, and I was thinking I might take those cases and bisect the lines into three vertices, and test those two lines until I don't have to give up.

GhostDaddio:
I'm a big fan of comments, unfortunately, I'm not so good at them.  As for the other forums, I go by BT there.  And, as I'm sure you've guessed, the whole mechanics of github are foreign to me.

[...]

PR #19, opened 2017-07-06
Limits

jeffeb3:
Limit vertices to stay inside the machine limits, but don't change the shape (when possible). 

I know this still has at least two issues:
 - When the limits change after the vertices, it's not updating the vertcies. I think there must be a simple way to fix this, but I'm not familiar with it.
 - When two consecutive points are out of bounds, like an entire side of a square, it's doing something funny, where it goes across the entire border. Needs some investigation.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -6,20 +6,48 @@ import Documentation from './Documentation';
 import Transforms from './inputs/Transforms';
 import MachinePreview from './MachinePreview';
 import GCodeGenerator from './GCode';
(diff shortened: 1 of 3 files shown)


jeffeb3:
w.r.t. #17

jeffeb3:
This really fails with the high resolution paths, like the logo. I made an assumption that if a point was oob, the next point wouldn't. That's no good, so I have to check for the case when next is also oob, and instead of moving towards it, I'll just add the nearestPoint.

PR #22, opened 2017-07-07
Draw the omage to vicious1.

jeffeb3:


jeffeb3:
Yeah, since I can't do arcs, it has quite a few points to define it. I'm assuming estlcam does something smart with resolution, but I want the shapes "normalized", but I can't drop the resolution, someone might still want to draw one 500mmx500mm.

PR #26, opened 2017-07-07
Fix calls to outOfBounds (((Why isn't this an exception?!))).

jeffeb3:


jeffeb3:
This isn't working right still. I am close with a more forceful solution.

PR #37, opened 2017-07-18
Limits 2

jeffeb3:
I think I have this thing nailed. This branch requires the 'victor' library, which seems like a reasonably detailed, MIT licensed, 2D Vector library:

http://victorjs.org/

It's not terrific, and I wish it had operators for things like addition, subtraction, etc. But I it's doing what it promises, and the price is right.


jeffeb3:
@jamesd256, Do you have any problems with me adding Victor as a dependency?

PR #40, opened 2017-07-20
Polygons

billsey:


jeffeb3:
The vertices in the reducer's shape list (state.shapes[the polygon].vertices) don't ever change. So even if you got the number of sides value back to the reducer, you won't be able to update those vertices.

Also, the number of sides, really only applies to the polygon. Even if you make your multipoint start shape, there will be different parameters. So we should probably move those inputs into panels on the shapes.

I think the best way to make these things dynamic is to replace the shape.vertices list with a function. So the reducer will call the function when it needs those vertices. The square, star, etc. will just return the prebuilt vertices, but the dynamic shapes will look at the state, and determine what their vertices should be.

So this:
```
  if (shape) {
    input = shape.vertices.map( (vertex) => {
      ...
    });
  }
```

Would be this:
```
  if (shape) {
    input = shape.vertices(state).map( (vertex) => {
      ...
    });
  }
```

And then Transforms.js would initialize the shapes with little functions that took the state and returned vertices, like this:
```
    this.props.addShape({
        name: "Square",
        vertices: (state) => {
          [
            Vertex(-1,-1),
            Vertex( 1,-1),
            Vertex( 1, 1),
            Vertex(-1, 1),
          ]},
      });
```

After you have that working, you will be passed in the state, and you can use something like state.polygonSides to compute the number of sides to use.

If you're still having fun, you can figure out how to do that. If you want me to just do it, I can get to it sometime this evening/night. Then you can just copy it for the multipoint star shape.

billsey:
I was thinking it would be pretty easy to just replace the triangle and square with calls to the polygon with appropriate number of sides. If you think of a star as two polygons with the same number of sides, but differing radius it'd be easy to create stars the same way, just change sides to points and add a percentage for determining the smaller radius as a proportion of the larger radius.
I'm in transit today and will be leaving next week for a couple of weeks sailing in Montana, but I should be able to play with it this weekend if you don't beat me to it.

jeffeb3:
I think I have this all merged in, with the inputs done and it has been really fun to play with. Thank you for writing this, and I'm sorry it's taken me 6 months to get around to it.

PR #42, opened 2017-08-08
Added an offset to the shapes.

jeffeb3:


jeffeb3:
My only complaint about this is that it's a little tricky to find a good combo, but I guess there were combinations that were crappy before as well.

PR #46, opened 2017-08-19
Wiper

jeffeb3:


jeffeb3:
It's a little rough. I think I would like to rest on this, and go through it again, cleaning up the names, and things. The work was a little trial and error, and it's overstepping at the ends, which causes some goofiness when I tried to do the offsets.

PR #94, opened 2020-01-25
move shapes into separate classes/files to make it easier to add and …

bobnik:
…remove them

As I was adding new shapes, I found it a bit perilous in that it required editing core files carefully. This PR attempts to make creating new shapes a more modular exercise. You define your shape in a custom class and then simply "register" it with Sandify. The core Sandify files can then remain untouched.

I am new to React/Redux, so hopefully my approach is architecturally-compliant.


src/reducers/Shapes.js
```diff
@@ -105,3 +47,5 @@ export default function shapes(state = defaultState, action) {
       return state;
   }
 }
+
+export default reduceReducers(defaultShapes, ...registeredShapes.map((shape) => shape.getReducer));
```
jeffeb3: This is neat! I didn't know about this.

src/machine/GCode.js
```diff
@@ -110,49 +110,16 @@ const getComments = (state) => {
 
   switch (state.app.input) {
     case 0: // shapes
-      var shapeName = state.shapes.shapes.find(shape => shape.name === state.shapes.currentShape).name;
+      var shapeInfo = state.shapes.shapes.find(shape => shape.name === state.shapes.currentShape);
 
       comments.push("  Content Type: Shapes");
       comments.push("    Starting Size: " + state.shapes.startingSize);
       comments.push("    Offset: X: " + state.transform.xformOffsetX + " Y: " + state.transform.xformOffsetY);
-      comments.push("    Selected Shape: " + shapeName);
```
jeffeb3: I'm very glad to see this go.

PR #97, opened 2020-03-06
More text

jeffeb3:
This is based on the font from Ray Solomon, which has been given to be incorporated into Sandify. To use it, select the text shape, and then select a font.

The monospace font is the older font.


jeffeb3:
@bobnik, This PR uses a dropdown box for the selection of a font. I probably could have done a little checkbox/toolbar thing, which may have been better for this situation. But this may be a good example for how we could do a shape dropdown, or a shape subsection dropdown. Let me know what you think (and also, if you can spend a few minutes taking a look and seeing if this all looks reasonable, I would appreciate it a lot).

jeffeb3:
@billsey, This is a PR with the two new fonts. I tried to not break the previous font, which is now the Monospace font.

src/inputs/Transforms.js
```diff
@@ -148,21 +151,44 @@ class Shape extends Component {
...
-                   onKeyDown={disableEnter}/>
-               </Col>
-             </FormGroup>
+      if (option.type && option.type === "dropdown") {
+        return <FormGroup controlId="options-step" key={option.title}>
+                 <Col componentClass={ControlLabel} sm={4}>
+                   {option.title}
+                 </Col>
+                 <Col sm={8}>
+                   <ButtonGroup justified>
```
bobnik: Personal preference, but I would remove the ButtonGroup here. The non-justified dropdown button looks better, less heavy, and left-aligned.

bobnik:
@jeffeb3 All looks good - made one small comment above. As far as a shape dropdown goes, I played around a bit and got sidelined by work. I hope to get back to it soon-ish.

PR #98, opened 2020-03-16
Feature/redux upgrade

bobnik:
This PR upgrades React and Redux, and does some pretty aggressive folder restructuring. Some details:

- Structure code base using a "features" pattern, where all files related to a given feature (e.g., machine) are grouped together. Vs separating files into concern-based folders (e.g., reducers)
- Use Redux Toolkit to reduce boilerplate. This involves creating "slices" that replace previous reducer and action type constant logic.
- Add specs for all feature reducers.
- Refactor shapes to be a bit more modular, and allow for reducer specs.
- There are development warnings coming from React that will likely go away when we upgrade to Bootstrap 4, which I'd like to tackle next.

I realize that this is a disruptive PR. The specs should hopefully mitigate the risk. I also did extensive UI manual testing. LMK if you have any questions!


jeffeb3:
I am getting an error while trying to test this version from my computer:
```
TypeError: Path must be a string. Received undefined
    at assertPath (path.js:28:11)
    at Object.join (path.js:1233:7)
    at noopServiceWorkerMiddleware (/home/jeffe/home/projects/sandify/node_modules/react-dev-utils/noopServiceWorkerMiddleware.js:14:26)
    at Layer.handle [as handle_request] (/home/jeffe/home/projects/sandify/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/home/jeffe/home/projects/sandify/node_modules/express/lib/router/index.js:317:13)
    at /home/jeffe/home/projects/sandify/node_modules/express/lib/router/index.js:284:7
    at Function.process_params (/home/jeffe/home/projects/sandify/node_modules/express/lib/router/index.js:335:12)
    at next (/home/jeffe/home/projects/sandify/node_modules/express/lib/router/index.js:275:10)
    at launchEditorMiddleware (/home/jeffe/home/projects/sandify/node_modules/react-dev-utils/errorOverlayMiddleware.js:20:7)
    at Layer.handle [as handle_request] (/home/jeffe/home/projects/sandify/node_modules/express/lib/router/layer.js:95:5)
```
I removed the npm_modules folder, and reinstalled the modules with `npm install`, but it didn't make a difference. I see one question about this related to a jest dependency:

https://stackoverflow.com/questions/46220674/jest-typeerror-path-must-be-a-string-received-undefined

But since Jest isn't in the stack trace, I'm not sure where this is coming from. Thoughts?

bobnik:
What version of react-dev-utils is locked? I had an issue at one point. Try
explicitly setting it to something lower than 10.1.2 (mine is locked at
10.1.0)

jeffeb3:
From the poking around I can do here in the diff, and in vim, I can see that this is a waaaay better. And the tests pass too. So that is great. Thank you a bunch for doing this. I am pretty sure I can find my way around in this code as good or better as I was, and I am sure it is easier for you, so as soon as I can verify this works just as well, I will merge this and run the deploy.

I am also tempted to get circleCI in here too, to run these tests, and then to deploy from master. It would make me feel better that nothing was coming from this machine and ending up deployed.

bobnik:
If that doesn’t work, I recommend going straight to the bootstrap PR and trying from there.

jeffeb3:
> What version of react-dev-utils is locked?
10.2.0. I am not a node expert, but that's what is in the package-lock.json.

bobnik:
Trying depending on 10.1.1 and rebuilding.

jeffeb3:
That did it. It is looking good.

PR #100, opened 2020-03-17
Feature/bootstrap upgrade

bobnik:
This is a continuation from the React/Redux upgrade. I'm submitting a PR now to give you the option of reviewing both together (and to save UI testing time). Included here:

- all of the React upgrades
- Bootstrap 4 migration
- some cosmetic adjustments/enhancements which are understandably subjective and can be reverted as need. Some of these changes were designed to move the UI toward future features like letting the user save/load shapes that each have their own distinct configuration.


bobnik:
@jeffeb3 This PR has the previous PR inside of it, but Github is not updating the visual diff. I _think_ this is okay, but it makes reviewing the specific changes difficult. This link seems to propose a way to fix it, but it looks like more trouble than it's worth. You may want to do a manual merge locally when you're ready just to be sure the merge goes without a hitch (vs doing it from Github). Anyway, LMK if you run into trouble when the time comes.

https://makandracards.com/makandra/13043-force-github-pull-requests-to-update-the-diff-against-its-target-branch

bobnik:
I have created a new, cleaner PR that doesn't have these issues.

PR #101, opened 2020-03-18
Feature/revised bootstrap upgrade

bobnik:
This is a continuation from the React/Redux upgrade. I'm submitting a PR now to give you the option of reviewing both together (and to save UI testing time). Included here:

- Bootstrap 4 migration
- some cosmetic adjustments/enhancements which are understandably subjective and can be reverted as need. Some of these changes were designed to move the UI toward future features like letting the user save/load shapes that each have their own distinct configuration.


bobnik:
@jeffeb3 I deleted the old PR. Please use this one when you review. It shows the proper diffs.

jeffeb3:
I have been looking at this, and I like it so much! I will merge and deploy this soon.

package.json
```diff
@@ -1,13 +1,14 @@
...
-  "version": "0.1.7",
+  "version": "0.1.8",
   "private": true,
   "dependencies": {
     "@reduxjs/toolkit": "^1.2.5",
+    "node-sass": "^4.13.1",
     "rc-slider": "^8.6.3",
     "react": "^16.13.0",
-    "react-bootstrap": "^0.31.0",
+    "react-bootstrap": "^1.0.0-beta.17",
```
jeffeb3: Should we be worried that this is marked beta?

src/features/transforms/Transforms.js
```diff
@@ -30,39 +34,78 @@ const mapDispatch = (dispatch, ownProps) => {
...
+          </Row>
+
+          <Row className="align-items-center pb-2">
+            <Col sm={4}>
+              <Form.Label htmlFor="shape-offset">
+              Offset
+              </Form.Label>
+            </Col>
+
+            <Col sm={8}>
```
jeffeb3: Something is funny with these items. They stretch over the right side when the screen is small (like on mobile). I can affect it a little by changing the sm, but that doesn't seem to be the right fix.

package.json
bobnik: I don't think so. It's a late beta, our UI is pretty minimal, and I didn't run into any issues. It also gets us to BS4, which is good for a number of reasons. That said, lmk if you see anything weird or feel this is risky.

src/features/transforms/Transforms.js
jeffeb3: screenshot-2020-03-18-1584570666

PR #102, opened 2020-03-27
Feature/shape configuration

bobnik:
As discussed, I'm sending across this PR so that you can play around. There are a few small issues that need resolving regardless, so please don't merge just yet.

- Ability to configure each shape separately.
- Swap position of preview window and tabs.

If we need to revert any major changes, I can branch from an earlier commit as needed to keep things clean and resubmit the PR.


jeffeb3:
screenshot-2020-03-28-1585451979

I have just a little time to check this out and run it. I really like the organization. The size of the machine preview should probably be smaller. It is really tall and has a scroll bar. The tab also has a scroll bar, and then the entire window has a scroll bar? Maybe the only issue is the large machine preview. 

I haven't had a chance to look at the code yet, which I want to really check out in detail.

bobnik:
You're right about the resizing. It was a bit off. I've made some changes.
You should only get a single vertical scrollbar in the right sidebar.
Everything else should autosize.

I've also added the ability to save user changes in local storage. Changes
can be reset by clicking "Restore defaults." I like the "stickiness" of
this, but if you don't like it, it's easy to revert.

jeffeb3:
The wiper, at least, is a pretty different shape type. It shouldn't have any offsets, or the ability to transform. I am wondering if it should have some different visual cue in the select list, so it stands out. It is a feature that was requested a few times, so I would hate for people to just not see it.

jeffeb3:
Does lodash/throttle need to be in the package.json?

jeffeb3:
There are a few scroll bars on some of the buttons. "Reverse path", "Rectangular machine", "Force Endpoints", "Spin". It looks like maybe they are 1pix too big.

jeffeb3:
I should have removed the turtle stuff a while ago. Sorry you had to drag that through this. We should remove that. 

In the future, I do imagine a script input, that will drive using the turtle language and then use the rest of sandify to generate gcode/thr from that, but that isn't a priority, especially as there are getting to be more interesting things to do without trying to teach people to code.

jeffeb3:
It looks like there is a startingSize in the shape, and in the transform. I think only the shape one is being used?

jeffeb3:
I'd like to bump the version, to at least 0.2.0. It would be nice if I had a way to publish this to a sidebar, and then let some people try it out. I can't break it, but it is so much different, I would feel better if there were just a few more eyes on it.

In general, this already passes the sniff test, and qualifies as "better". That is the ultimate line that needs to be crossed in a PR. I am only hoping I can get a little time soon to address some of this, before merging and deploying. But soon.

bobnik:
> Does lodash/throttle need to be in the package.json?

Yes. Ir's used to throttle preview window redrawing when the user resizes the window. Otherwise, it's redrawing constantly and unnecessarily. I was also seeing react errors if I resized quickly/often.

bobnik:
> It looks like there is a startingSize in the shape, and in the transform. I think only the shape one is being used?

You're right. It can be removed from transform.

bobnik:
> There are a few scroll bars on some of the buttons. "Reverse path", "Rectangular machine", "Force Endpoints", "Spin". It looks like maybe they are 1pix too big.

I see it in Firefox, but not Chrome. Weird. Anyway, I've pushed a fix for this.

bobnik:
> The wiper, at least, is a pretty different shape type. It shouldn't have any offsets, or the ability to transform. I am wondering if it should have some different visual cue in the select list, so it stands out. It is a feature that was requested a few times, so I would hate for people to just not see it.

I've pushed a commit to prevent wiper from being transformed. As far as making it more prominent, what if we moved it up in the drop list order? If that's not enough, we could (1) change out the drop list with an inline list, or (2) find a way to style the wiper element in the drop list. Reluctant to do #1 for real estate reasons.

jeffeb3:
> > Does lodash/throttle need to be in the package.json?
> 
> Yes. Ir's used to throttle preview window redrawing when the user resizes the window. Otherwise, it's redrawing constantly and unnecessarily. I was also seeing react errors if I resized quickly/often.

Right. But it isn't in package.json. Unless it is included in one of these other packages.

bobnik:
> > > Does lodash/throttle need to be in the package.json?
> > 
> > 
> > Yes. Ir's used to throttle preview window redrawing when the user resizes the window. Otherwise, it's redrawing constantly and unnecessarily. I was also seeing react errors if I resized quickly/often.
> 
> Right. But it isn't in package.json. Unless it is included in one of these other packages.

Got it. It's required by several other packages, but I pushed a commit to explicitly include it, since we're using it.

jeffeb3:
> > The wiper, at least, is a pretty different shape type. It shouldn't have any offsets, or the ability to transform. I am wondering if it should have some different visual cue in the select list, so it stands out. It is a feature that was requested a few times, so I would hate for people to just not see it.
> 
> I've pushed a commit to prevent wiper from being transformed. As far as making it more prominent, what if we moved it up in the drop list order? If that's not enough, we could (1) change out the drop list with an inline list, or (2) find a way to style the wiper element in the drop list. Reluctant to do #1 for real estate reasons.

I put it at the bottom of the list, and I made it part of it's own group. A shape can specify it's "selectGroup" and then the shapesList will group them together in the Select. It feels a little redundant to say "Shapes" and "Wiper" again, but I'm ok with that. If there are any other ways you think we should sort or group those, I'm all ears.

bobnik:
> I put it at the bottom of the list, and I made it part of it's own group. A shape can specify it's "selectGroup" and then the shapesList will group them together in the Select. It feels a little redundant to say "Shapes" and "Wiper" again, but I'm ok with that. If there are any other ways you think we should sort or group those, I'm all ears.
This looks good. Low priority: Maybe call the group "erasers" or "erase"? That's what the track type is called in Sisyphus, and we could add other erasers later.

jeffeb3:
I'm going to merge this. Then I'll go look at the design/shape stuff you sent me. I think it makes sense to wait to deploy until we get those new shapes in, and I want to try to make the text look a little better, and maybe try some of the other text features I want (like multiple lines).

PR #105, opened 2020-04-04
Feature/tessellation and fractal spirograph

bobnik:
This PR has two new shapes - fractal spirograph and tessellation twist. It also makes style changes to the track colors to better indicate direction. Finally, it uses min/max to prevent user from specifying bad shape values.

### DIFF
--- a/src/common/Graph.js
+++ b/src/common/Graph.js
@@ -1,0 +1,77 @@
+import { PriorityQueue } from './PriorityQueue.js'
+
+export const mix = (u, v, s) => {
+  var result = []
+  for (var i = 0; i < u.length; ++i ) {
+    result.push((1.0 - s) * u[i] + s * v[i])
+  }
+
+  return result
+}
+
+export const vec2 = (...args) => {
+  switch (args.length) {
+    case 0: args.push(0.0); break
+    default: args.push(0.0)
+  }
+
+  return args.splice(0, 2)
+}
+
+export class Graph {
+  constructor() {
+    this.nodes = []
+    this.adjacencyList = {}
+    this.edges = []
+  }
+
+  addNode(node) {
+    this.nodes.push(node)
+    this.adjacencyList[node] = []
+  }
+
+  addEdge(node1, node2, weight=1) {
+    this.adjacencyList[node1].push({node:node2, weight: weight})
+    this.adjacencyList[node2].push({node:node1, weight: weight})
+    this.edges.push([node1.toString(), node2.toString()].sort().toString())
+  }
+
+  dijkstraShortestPath(startNode, endNode) {
+    let times = {}
+    let backtrace = {}
+    let pq = new PriorityQueue()
+
+    times[startNode] = 0
+
+    this.nodes.forEach(node => {
+      if (node !== startNode) {
+        times[node] = Infinity
+      }
+    })
+
+    pq.enqueue([startNode, 0])
+
+    while (!pq.isEmpty()) {
+      let shortestStep = pq.dequeue()
+      let currentNode = shortestStep[0]
+      this.adjacencyList[currentNode].forEach(neighbor => {
+        let time = times[currentNode] + neighbor.weight
+
+        if (time < times[neighbor.node]) {
+          times[neighbor.node] = time
+          backtrace[neighbor.node] = currentNode
+          pq.enqueue([neighbor.node, time])
+        }
+      })
+    }
+
+    let path = [endNode]
+    let lastStep = endNode
+    while(lastStep !== startNode) {
+      path.unshift(backtrace[lastStep])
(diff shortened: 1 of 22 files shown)


src/common/Graph.js
```diff
@@ -0,0 +1,77 @@
...
+    this.adjacencyList[node] = []
+  }
+
+  addEdge(node1, node2, weight=1) {
+    this.adjacencyList[node1].push({node:node2, weight: weight})
+    this.adjacencyList[node2].push({node:node1, weight: weight})
+    this.edges.push([node1.toString(), node2.toString()].sort().toString())
+  }
+
+  dijkstraShortestPath(startNode, endNode) {
```
jeffeb3, line 39: This is stepping up a notch!

src/components/CheckboxOption.js
```diff
@@ -0,0 +1,37 @@
...
+    return (
+      <Row className="align-items-center pb-2">
+        <Col sm={5}>
+        <Form.Label htmlFor="options-step">
+          {option.title}
+        </Form.Label>
+        </Col>
+
+        <Col sm={7}>
+          <Form.Check
```
jeffeb3, line 22: The other boolean options (like reverse path) are big toggle buttons. What about using one of those instead?

src/features/app/store.js
```diff
@@ -35,8 +33,41 @@ Object.keys(registeredShapes).forEach(key => {
   store.dispatch(addShape(state))
 })
 
+// set to true when running locally if you want to preserve your shape
+// settings across page loads; don't forget to toggle false when done testing!
+const persistState = false
```
jeffeb3, line 38: It is definitely helpful to use when running locally. For sure.

src/shapes/fractal_spirograph/FractalSpirograph.js
```diff
@@ -0,0 +1,95 @@
...
+    this.link = 'https://benice-equation.blogspot.com/2012/01/fractal-spirograph.html'
+  }
+
+  getInitialState() {
+    return {
+      ...super.getInitialState(),
+      ...{
+        type: 'fractal_spirograph',
+        repeatEnabled: false,
+        startingSize: 54,
```
jeffeb3, line 46: This is all very nice. Very simple and clear. :+1:

src/shapes/tessellation_twist/TessellationTwist.js
```diff
@@ -0,0 +1,266 @@
+import Victor from 'victor'
+import { Graph, vec2, mix } from '../../common/Graph'
+import eulerianTrail from '../../common/eulerianTrail'
+import { difference } from '../../common/util'
+import Shape, { shapeOptions } from '../Shape'
+
+const vec_triangle = [
```
jeffeb3, line 7: These vertices seem a lot like the math I worked out for the releaux triangle. Although that was the center of the arc circles, so it is rotated a bit. IDK if higher number of sides is interesting to you, but if this is the limit, we could probably generalize this for almost any number of sides (greater than three).

src/shapes/tessellation_twist/TessellationTwist.js
```diff
@@ -0,0 +1,266 @@
...
+      min: 0
+    }
+  }
+}
+
+// Adapted from https://codepen.io/rafaelpascoalrodrigues/pen/KpBJve.
+export default class TessellationTwist extends Shape {
+  constructor() {
+    super('Tessellation Twist')
+  }
```
jeffeb3, line 134: Is there a link for this one? It looks like there is a lot of interesting math, including some travelling salesman solving?

src/shapes/tessellation_twist/TessellationTwist.js
```diff
@@ -0,0 +1,266 @@
...
+
+    trail.forEach((key, index) => {
+      let vertex = vertexMap.get(key)
+
+      if (prevKey) {
+        let edgeKey = [key, prevKey].sort().toString()
+
+        if (!edgeMap.get(edgeKey)) {
+          // non-eulerian move, so we'll walk the shortest valid path between them
+          let path = graph.dijkstraShortestPath(prevKey, key)
```
jeffeb3, line 230: :+1:

src/components/CheckboxOption.js
bobnik: I feel that the toggle button should be used only for accordion-reveals of more options. Otherwise it's too heavyweight and disrupts the visual layout of the properties. I like having an "inline", lighter option. That said, I can certainly convert this option to a toggle button.

src/shapes/tessellation_twist/TessellationTwist.js
bobnik: That would be cool. I don't think I could work out the math myself though :)

PR #107, opened 2020-04-05
Feature/start end points

bobnik:
With Sisyphus, it's important to be able to control the starting and ending points separately. Sisyphus recommends that the start and end points must be either r=0, or r=1 (radius). It uses that information within a playlist of multiple patterns to determine whether it needs to "home" (move to center or perimeter before starting the next pattern). If a pattern follows one of the [0-0, 0-1, 1-0, 1-1] patterns, it can also auto-reverse a pattern before starting it to avoid having to home, which creates unpleasing lines.

This PR allows the user to control both the start and the end point of the shape. In the context of a playlist, you can then carefully set these depending on the specific pattern and its position in the playlist to avoid homing altogether.

- add start point and end point to polar machine settings, removing the force endpoints toggle
- minor refactoring to make it easier to add new machine setting


src/features/machine/PolarSettings.js
```diff
@@ -61,14 +65,37 @@ class PolarSettings extends Component {
...
+            <Row className="align-items-center pb-2">
+              <Col sm={4}>
+                <Form.Label htmlFor="forceStart">
+                  Start point
+                </Form.Label>
+              </Col>
+
+              <Col sm={8}>
+                <ToggleButtonGroup id="forceStart" type="radio" name="origin" value={this.props.startPoint} onChange={this.props.onStartPointChange}>
+                  <ToggleButton variant="light" value={0}>auto</ToggleButton>
```
jeffeb3, line 77: I think "auto" may not be the best world. Maybe "none" instead?

jeffeb3, line 77: 0,1,2 is fine, but we could easily do more descriptive things, right? Would a string be the right choice, or is there some kind of enum?

jeffeb3: The concept of auto is actually what I was hoping for. Something where the start and end points would would be determined by how far away the pattern's start and end was from the center.

src/common/Computer.js
```diff
@@ -75,129 +75,137 @@ export const transform = (data, vertex, fraction_index) => {
...
+  while (nextCorner !== machine.rectOrigin[0]) {
+    console.log("nextCorner: " + nextCorner)
+    newVertices.push({ ...first, x: corners[nextCorner].x, y: corners[nextCorner].y})
+    nextCorner -= 1
+    if (nextCorner < 0) {
+      nextCorner = 3
+    }
+  }
+
+  console.log("nextCorner: " + nextCorner)
```
jeffeb3, line 145: These consoles are my fault, but they shouldn't be in here anymore...

jeffeb3: I can fix them when I get back to a computer.

src/features/machine/PolarSettings.js
jeffeb3: We could also make them toggle, so if you had auto it would pick, but if you selected none of them, it wouldn't put the start or end in.

bobnik: I think string (enum) is a good approach. And I'm not sure auto is what you want in the context of a playlist - the user needs to decide. I'll make the changes soon.

PR #115, opened 2020-04-06
don't limit the number of tessellation polygon sides

bobnik:
@jeffeb3 per your comment in #105, I've refactored the tessellation to allow any number of polygon sides. I like it - you can create nice circular patterns now for polar machines.


jeffeb3:
LGTM

PR #118, opened 2020-04-09
Feature/toggle switch

bobnik:
As discussed, swaps out checkbox and/or accordion toggle for a toggle switch where appropriate. 

I also made a few minor style changes: Rectangle force origin toggle buttons are displayed in 2x2 grid, and theta-rho tab does not display file metadata unless a file has been imported.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,8 @@
   "private": true,
   "dependencies": {
     "@reduxjs/toolkit": "^1.2.5",
-    "lodash": "^4.17.15",
     "color": "^3.1.2",
+    "lodash": "^4.17.15",
     "node-sass": "^4.13.1",
     "rc-slider": "^8.6.3",
     "react": "^16.13.0",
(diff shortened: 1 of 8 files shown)


jeffeb3:
If there are no comments in a .thr file, then the gray details block isn't shown. I changed it to be gated on the presence of a file name.

jeffeb3:
Something else is funny though. It is really slowing down. The thr input has always been a bit slower, because it is sampling the arcs at a pretty high resolution to keep their round shape. But when I load up a pretty simple file, the CPU usage goes way up, and it takes several seconds to change something like the zoom. I will email you the test file I tried.

jeffeb3:
These toggles look really good though.

> Something else is funny though. It is really slowing down. The thr input has always been a bit slower, because it is sampling the arcs at a pretty high resolution to keep their round shape. But when I load up a pretty simple file, the CPU usage goes way up, and it takes several seconds to change something like the zoom. I will email you the test file I tried.

I could not repeat this issue. I think it might have been something local to this machine.

PR #119, opened 2020-04-12
Features/text formatting

jeffeb3:
I am trying to solve three main problems here:
1. The most obvious is that I have a textarea now, and multiple lines of text
2. I added a rotate feature, and it can handle scrolling text around in rings. It will manage text you read mostly from the top and text you read mostly from the bottom
3. The width and spacing of the two Ray fonts are still not as good as they are in the other tool. I managed to find a 2.0*x, and that helped. The sans serif font also has a variable distance between certain letters (Kern?). I adding in the generic spacing, but I need to still work out a way to make the spacing dependent on the previous and next letter.

This is working ok. I need to get it on the machine, and actually see how it looks. My machine is currently out of service.

The code could be cleaner, especially the math in the rotate section, which was made sort of my trial and error.

screenshot-2020-04-11-1586670678

screenshot-2020-04-11-1586670696


src/shapes/input_text/InputText.js
```diff
@@ -14,9 +14,25 @@ const options = {
       type: 'dropdown',
       choices: ['Cursive', 'Sans Serif', 'Monospace'],
     },
+    rotateDir: {
+      title: 'Rotate',
+      type: 'dropdown',
+      choices: ['Top', 'Center', 'Bottom'],
```
bobnik, line 20: I think these should be lowercase for consistency.

src/shapes/input_text/InputText.js
```diff
@@ -39,40 +56,170 @@ export default class InputText extends Shape {
...
+      let y = (lines.length - 1) * maxY / 2.0
+      let textPoints = []
+
+      // Capture some wrap around points, to connect the lines.
+      let connectorPoints = []
+      lines.forEach( (points) => {
+        let maxX = getMaxX(points)
+        let widthOffset = maxX / 2.0
+
+        // Add in the connector points (if we have any)
```
bobnik, line 117: Not sure this is the relevant line of code, but... can you use traceCircle here? If I recall correctly, it has the added benefit of traveling clockwise or counterclockwise to minimize the distance. I think the text will look better when rotate = "top" or "bottom" if you don't travel most of the connecting circle each time.

src/shapes/input_text/InputText.js
bobnik, line 20: image

Do you think for bottom text, you should reverse the lines of text so that reading from top to bottom is the same?

jeffeb3: Great idea.

src/shapes/input_text/InputText.js
jeffeb3: Great idea.

src/shapes/input_text/InputText.js
jeffeb3: This controls the text in the options, so the user sees them. I think Capital makes more sense for that. The fonts are also Capitalized. I know the shape names are lowercase, but they aren't seen in the UI.

Issue #73, opened 2019-02-15
gcode optimization


mrehorst:
I've started working on a flow chart for a post processor for optimizing gcode.  My goal is to eliminate duplicate lines in the gcode files, but more importantly, to minimize the motion around the edges of the table while still preserving the sequence of the pattern generation.  The motion around the edges is boring to watch, takes up a lot of time, and adds to wear on the mechanism.

I've been making sketches and looking at patterns for a couple days and think I have a handle on all the special cases that have to be allowed:

Stuff to keep from the source file:
1) any non duplicate, non G01 line
2) the first G01 line
3) all non-duplicate, on-table G01 lines
4) edge to table transitions
5) table to edge transitions
6) corner to corner transitions
7) edge to corner transitions
8) corner to edge transitions
9) corner to table transitions 
10) table to corner transitions
11) edge to other edge transitions

Stuff to throw away:
1) sequential duplicate lines
2) all "in between" points along an edge
3) in between points along multiple edges

Regarding transitions-  the key point is that you have to keep anything that draws on the table, and minimize the travel between the points where the ball leaves the table (a transition) and where it reenters the table (another transition).  The source file may have the ball racing back and forth on the same edge, or around 3 corners several times before it gets back onto the table.  We can throw away all the in-between stuff and just keep (or generate) a (the shortest?) path between the ball exiting and entering the table along as many edges as it takes.

I'm trying to figure out how to make this work in a single pass, reading the source gcode line by line.  I suspect it may be easier to use multiple passes, (example- one pass that eliminates duplicate lines is very easy), but we'll see.

This is going to take a while for me to finish because I am going to have to learn whatever programming language I end up choosing once I have worked out the algorithm.  I may be able to get some coding help from people at the makerspace.

jeffeb3:
It would be really really useful to do this in the javascript. Having a python, perl, basic whatever tool the user has to use later is going to really limit the utility. If you are just making this for yourself, then that's fine, but putting it in the website would make it instantly useable by everyone.

Removing duplicates is super easy, and we can do that with about 2 lines of code in the gcode.js file.

It sounds like the rest is:

1) Watch for a transition to the edge.
2) scout ahead for the transition back from the edge.
3) replace all of that with the shortest path from transitions.

The challenges will be: a) detecting transitions, b) the data structures and managing the points safely and c) creating the shortest path (which is always harder than I think).

mrehorst:
[...]

PR #121, opened 2020-04-15
Feature/optimize perimeter

bobnik:
This PR tries to implement perimeter optimization similar in concept to what mrehorst suggested in   #73. I implemented this initially for the polar machine before I even saw Mark’s issue. Specifically, this PR tries to:

- Eliminates unnecessary perimeter moves. This reduces the number of moves considerably in some cases, and when you increase the size or number of loops outside of permitted boundaries, the total distance no longer increases indefinitely, instead hitting a local maximum.
- If ‘minimizeMoves’ is enabled, tries to reduce the total number of perimeter moves by walking “perimeter segments” efficiently. I added this as a toggle-able option instead of an automatic improvement, because it changes the path (not always as symmetrically pleasing to the eye) and sometimes it does not actually reduce the total distance. That said, more often than not, it looks pleasing and reduces the distance considerably. For example, try Clover, small circle radius = 1.85, number of loops = 17 as example. Toggle ‘minimize perimeter moves’ and use the slider to see the difference.
- Moves limit enforcement and end point logic into machine classes. I was surprised at how much of the overall logic could be abstracted into a superclass and applied to both rectangular and polar movement.

Performance considerations: This PR does introduce some additional calculations with each redraw. I did not time them to see the difference. On my machine (high-end), the additional calculation load did not affect the responsiveness of the UI even when sliding. That said, if you feel this impacts performance there may be ways to offload some of the calculations to a background worker or throttle differently, etc.


bobnik:
Something went wrong with this PR... going to try again.

PR #122, opened 2020-04-15
Feature/optimize perimeter

bobnik:
This PR tries to implement perimeter optimization similar in concept to what mrehorst suggested in   #73. I implemented this initially for the polar machine before I even saw Mark’s issue. Specifically, this PR tries to:

- Eliminates unnecessary perimeter moves. This reduces the number of moves considerably in some cases, and when you increase the size or number of loops outside of permitted boundaries, the total distance no longer increases indefinitely, instead hitting a local maximum.
- If ‘minimizeMoves’ is enabled, tries to reduce the total number of perimeter moves by walking “perimeter segments” efficiently. I added this as a toggle-able option instead of an automatic improvement, because it changes the path (not always as symmetrically pleasing to the eye) and sometimes it does not actually reduce the total distance. That said, more often than not, it looks pleasing and reduces the distance considerably. For example, try Clover, small circle radius = 1.85, number of loops = 17 as example. Toggle ‘minimize perimeter moves’ and use the slider to see the difference.
- Moves limit enforcement and end point logic into machine classes. I was pleased at how much of the overall logic could be abstracted into a superclass and applied to both rectangular and polar movement.

Performance considerations: This PR does introduce some additional calculations with each redraw. I did not time them to see the difference. On my machine (high-end), the additional calculation load did not affect the responsiveness of the UI even when sliding. That said, if you feel this impacts performance there may be ways to offload some of the calculations to a background worker or throttle differently, etc.


bobnik:
image

Example I am running on my table now: with PR changes, distance of 66k; without 150k

src/features/gcode/GCodeGenerator.js
```diff
@@ -184,7 +183,7 @@ class GCodeGenerator extends Component {
 
         // This loads up (start, end].
         for (let step = 0; step < (delta.magnitude() / maxLength) ; step++) {
-          subsampledVertices.push(Vertex(start.x + step * deltaSegment.x,
+          subsampledVertices.push(new Victor(start.x + step * deltaSegment.x,
                                          start.y + step * deltaSegment.y,
                                          this.props.vertices[next].f))
```
jeffeb3, line 188: I think we can drop these f variables.

bobnik: ah, yes. I missed this one (and maybe others) in the Victor replacement.

src/features/gcode/GCodeGenerator.js
```diff
@@ -224,7 +223,7 @@ class GCodeGenerator extends Component {
       previousRawTheta = rawTheta
       previousTheta = theta
 
-      trVertices.push(Vertex(theta, rho, subsampledVertices[next].f))
```
jeffeb3, line 227: Here is another one.

src/features/machine/PreviewWindow.js
```diff
@@ -159,32 +159,39 @@ class PreviewWindow extends Component {
...
-      let drawing_vertices = this.slice_vertices(this.props.vertices, this.props.sliderValue)
+      let sliderRange = this.sliderVertexRange(this.props.vertices, this.props.sliderValue)
+      let drawing_vertices = this.props.vertices.slice(sliderRange[0], sliderRange[1] + 1)
 
       // Draw the background vertices
       if (this.props.sliderValue !== 0) {
         context.beginPath()
         context.lineWidth = this.mmToPixelsScale()
-        context.strokeStyle = "rgba(204, 204, 204, 0.35)"
+        context.strokeStyle = Color('#6E6E00')
```
jeffeb3, line 186: I was skeptical, but I really like how this ends up looking. It still gives me a sense of start to end, but now I also don't have to remember which points are in the past vs. future.

src/features/machine/computer.js
```diff
@@ -232,21 +130,11 @@ function buildTrackLoop(state, i, t) {
...
-      }
-
-      const sizeX = (machine.maxX - machine.minX)/2.0
-      const sizeY = (machine.maxY - machine.minY)/2.0
-      vertices = enforceRectLimits(vertices, sizeX, sizeY)
-    } else {
-      vertices = addPolarEndpoints(machine, vertices)
-      vertices = enforcePolarLimits(vertices, machine.maxRadius)
-    }
+    vertices = new machineClass(vertices, state.machine).polish().vertices
```
jeffeb3, line 137: Is this common in javascript? Just make a temporary class, call the interesting function on it, grab what you want out of it, and throw it in the garbage :). I like it.

src/features/machine/machineSlice.js
```diff
@@ -56,6 +57,10 @@ const machineSlice = createSlice({
       }
       state.rectOrigin = newValue
     },
+    toggleMinimizeMoves(state, action) {
+      state.minimizeMoves = !state.minimizeMoves
+      localStorage.setItem('minimizeMoves', state.minimizeMoves)
```
jeffeb3, line 62: From the user perspective, does minimize moves make more sense as a toggle in the output dialog? Similar to the reverse toggle?

src/features/machine/polarMachine.js
```diff
@@ -0,0 +1,210 @@
+import { angle, onSegment } from '../../common/Geometry'
+import Machine from './machine'
+import Victor from 'victor'
+
+export const traceCircle = (start, end, size) => {
```
jeffeb3, line 5: I had to change the function signature, since I didn't have an object with a .angle in my code. I will push a fix.

src/features/machine/polarMachine.js
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  // Returns whether a given path lies on the perimeter of the circle.
+  onPerimeter(v1, v2, delta=.001) {
+    let rm = Math.pow(this.settings.maxRadius, 2)
+    let r1 = Math.pow(v1.x, 2) + Math.pow(v1.y, 2)
+    let r2 = Math.pow(v2.x, 2) + Math.pow(v2.y, 2)
+    let d = this.perimeterDistance(v1, v2)
+
+    return (r1 >= rm - delta && r2 >= rm - delta) && d < 15
```
jeffeb3, line 99: Where does the 15 come from? It shouldn't ever get triggered, right? The d is in radians, [-pi,pi]?

src/features/machine/rectMachine.js
```diff
@@ -0,0 +1,333 @@
...
+        // It's going outward
+        scale = maxRadius / last.magnitude()
+        outPoint = last
+      } else {
+        scale = maxRadius / first.magnitude()
+        outPoint = first
+      }
+
+      let clipped = this.clipLine(
+        this.vertices[this.vertices.length - 1],
```
jeffeb3, line 44: We just figured out if we were growing or shrinking, and used that to find the ideal perimeter point, but then we always add this to the last point. I think adding it to the last point makes the most sense. So I'm not sure why I was checking for growing or shrinking.

src/features/machine/rectMachine.js
```diff
@@ -0,0 +1,333 @@
...
+    return this.distance(this.tracePerimeter(v1, v2, true))
+  }
+
+  // Returns whether a given path lies on the perimeter of the rectangle
+  onPerimeter(v1, v2, delta=.00001) {
+    const dx = Math.abs(Math.abs(v1.x) - this.sizeX)
+    const dy = Math.abs(Math.abs(v1.y) - this.sizeY)
+
+    if (dx < delta || dy < delta) {
+      return v1.x === v2.x || v1.y === v2.y
```
jeffeb3, line 65: This would still return true if dx < delta, and v1.y === v2.y... I think that's wrong. So a line from -250,100 to 50,100 wouldn't be on the perimeter.

src/features/machine/rectMachine.js
```diff
@@ -0,0 +1,333 @@
...
+  }
+
+  // Given two perimeter points, traces the shortest valid path between them (stays on
+  // perimeter). Returns a list of intermediate points on that path (if any).
+  // On further consideration, this could be redone using Dijsktra's algorithm, I believe,
+  // but this works and is, I believe, reasonably efficient.
+  tracePerimeter(p1, p2, includeOriginalPoints=false) {
+    let points
+
+    if ((p1.x === p2.x && Math.abs(p1.x) === this.sizeX) || (p1.y === p2.y && (Math.abs(p1.y) === this.sizeY))) {
```
jeffeb3, line 78: This looks like the perfect onPerimeter function, right? Except maybe this doesn't include a delta argument?

src/features/machine/rectMachine.js
```diff
@@ -0,0 +1,333 @@
...
+      // on the same line; no connecting points needed
+      points = []
+    } else {
+      // horizontal or vertical orientation
+      let o1 = Math.abs(p1.x) === this.sizeX ? 'v' : 'h'
+      let o2 = Math.abs(p2.x) === this.sizeX ? 'v' : 'h'
+
+      if (o1 !== o2) {
+        // connects via a single corner
+        points = (o1 === 'h') ?
```
jeffeb3, line 88: This is some very clever stuff. :100:

src/features/machine/rectMachine.spec.js
```diff
@@ -0,0 +1,32 @@
+import RectMachine from './rectMachine'
+
+describe('rect machine', () => {
```
jeffeb3, line 3: Yes. Excellent.

src/features/machine/rectMachine.js
jeffeb3: I updated the tests and fixed this function (I hope).

src/features/machine/rectMachine.js
jeffeb3: It looks funny if I always add it to the last point, instead of the outermost point. So I changed this around to insert the points to the inside if the pattern is shrinking.

jeffeb3:
I merged in master. I didn't do it with a rebase, or even a --no-ff.

jeffeb3:
I am totally fine with this though. It passes the bar for me. I will wait to merge, until I get a thumbs up from you. In case you disagree with my changes.

src/features/machine/machineSlice.js
bobnik: I don't think so, because you really need to see what it looks like before you output it. I've generally found that any pattern that has overlapping lines will not look good minimized because you clobber your elegant pattern when you draw over it out of sequence. Very occasionally, it actually _adds_ distance. I'm sure that could be fixed, but I concluded for now anyway that it wasn't worth changing.

bobnik:
> I am totally fine with this though. It passes the bar for me. I will wait to merge, until I get a thumbs up from you. In case you disagree with my changes.

The changes look good. I think when you merged in the latest changes from master, however, we somehow ended up with double commits. I created a new PR #124 to clean this up, so please merge that one instead.

jeffeb3:
Those extra commits came in because I did a git merge, and I didn't do a git merge --no-ff. So they were able to just be fast forwarded in. They wouldn't have made a difference to the code, but rebasing is cleaner. Thanks.

PR #123, opened 2020-04-15
npm test ran in github.

jeffeb3:


jeffeb3:
This is a work in progress, but I have to have a PR to really test this.

Allted:
Busy beaver.....getting exciting

jeffeb3:
OK. No longer a WIP. I think this is actually working to do this:
- run npm test, and only succeed if those tests pass. All PRs should be tested after this gets to master.
- run the test with npm versions 8, 10, 12. The deploy will be in 12, so that is probably what I should install locally.
- If I make a release, it triggers the Deploy workflow. I have only tested this so far with a dummy deploy step, but I think this will work when we are ready to deploy 0.2.1. I fully expect to need to fix something when I try to make a release, and it fails to run.

jeffeb3:
Sorry for the extra tags and releases. I had to make a release to test the action.

PR #125, opened 2020-04-17
Feature/file capitalization

bobnik:
This is a tiny PR to get a few files capitalized correctly. Awkward rename commits due to fact that Git does not recognize capitalization changes due MacOS file system case insensitivity.


jeffeb3:
I guess 'push' is really just things that get pushed in the jeffeb3 repo? And pull_request is needed to include things that are PR'ed from bobnik?

At any rate, this is the same errors I'm seeing here (I run ubuntu, which is very case sensitive).

bobnik:
> I guess 'push' is really just things that get pushed in the jeffeb3 repo? And pull_request is needed to include things that are PR'ed from bobnik?
> 
> At any rate, this is the same errors I'm seeing here (I run ubuntu, which is very case sensitive).

I found the issue. Gimme a sec. Sorry, Mac didn't flag this.

Issue #113, opened 2020-04-05
Import patterns aren't sticking.


jeffeb3:
When I open a .thr file, and then go to the machine tab, the vectors generated are from the shapes tab, not the most recently used tab.

bobnik:
@jeffeb3, if I understand you, I can see that's confusing, but I'm not sure the answer is to remember the last tab visited before previewing. I think that a successful .thr import should add a shape to the drop list (with configuration loaded), and change tabs to Shapes. This, of course, would require we fully support import of configuration values.

jeffeb3:
The import of thr files isn't for thr files created by sandify, it is for community generated thr files. It is mostly useful for me to draw these great thr files on my corexy machine. I also imagined it would be useful if someone wanted to scale or offset a thr file and export a thr file.

You should be able to import a thr file, then adjust your machine settings and export it without touching any of the other shapes.

I can see you saying, "We should move the import into the shape dropdown, like the wiper". Well, theoretical Bob, I think that is a good idea!

Issue #62, opened 2018-11-22
Import gcode files


jeffeb3:
Similar to the way the thetarho import lets gcode users convert thr files, thr users should be able to convert gcode files.

This is a little more complicated, because I'll need to parse the gcode for the features, which will mean at least supporting everything G0/1/2/3 does. But I'd like to do that anyway. Maybe someone knows of a good gcode javascript library?

In the UI, it probably makes sense to have it in the Code tab, and have the import button handle both, with some auto detection based on the extension, or the content. The rest shouldn't change.

billsey:
No idea how good they are, but I see several on https://libraries.io/search?keywords=gcode&languages=JavaScript

jeffeb3:
Thanks Bill! This one looks pretty good (MIT, and it's part of cnc.js): https://github.com/cncjs/gcode-toolpath/blob/master/README.md

PR #126, opened 2020-04-19
Feature/gcode input

jeffeb3:
### GCode Input

Similar to the thr input, this will import a gcode file and use the vertices as the source to generate either a gcode or thr output.

### Motivation

The main benefit of this (for thr and gcode users) is to be able to bring in, and somewhat manipulate the patterns generated for gcode. They can be made in inkscape, for a plotter, with CAM software, or even sandify.

### TODO
x The language on the Import page focuses on .thr files. I think the copyright info is good, and I hope we can keep it. I do not know of any good repositories for gcode patterns, but I am happy to link those there too.
x More importantly, the discussion in #113 about where the input pattern lives would be good to get resolved.

### Verification
There is a .gcode file I put in the pattern_input folder. It has some lines and arcs. It looks good, but was manually generated, so of course, I did the arcs the way I thought they should be done. But then I grabbed an arc torture test from the v1engineering forums, and it turned out correct looking too.

This solves issue #62


jeffeb3:
I solved #113 in this PR by looking at the state.file.name in the machine and about tabs. If it isn't empty, then I use the vertices from the import. If it is, then I use the transformVertices, from the shapes. That doesn't have the stickyness of the "last tab" hack, but it doesn't give you a way to easily clear the state if you've imported something either. So we may have to revisit this in the future.

I think this is ready to go.

jeffeb3:
Rebased.

jeffeb3:
@bobnik Do you have any thoughts on this PR?

bobnik:
> @bobnik Do you have any thoughts on this PR?

Looks good to me. I tried 5 different random files I found on the web, and they all imported well. As mentioned, I don't know much about this format.

PR #127, opened 2020-04-19
Feature/space filling curve eraser

bobnik:
This PR introduces a series of space filling curves under a single "Space Filler" eraser shape. Each curve does not actually render as a curve, as you'll see. 

I also added a few input callbacks, similar to what you (Jeff) proposed awhile back. You can configure a shape with callbacks for min, max, isVisible, and onChange. This gives us a fair amount of flexibility on configuring field options.

The underlying "machine" for these curves is an L-system rewrite engine. See https://en.wikipedia.org/wiki/L-system. I ported a simplistic and limited L-system, but I'm hoping in subsequent PRs to move to a more complex one that will let us create some amazing fractal and other patterns. We could even let users code their own (think of it like a limited turtle language), but it's pretty complicated. Anyway, expect to see some more cool L-system shapes soon!

image

### DIFF
--- a/src/common/geometry.js
+++ b/src/common/geometry.js
@@ -1,3 +1,5 @@
+import Victor from 'victor'
+
 // convert degrees to radians
 export const degToRad = (deg) => {
   return deg / 180.0 * Math.PI
@@ -30,4 +32,42 @@ export const slope = (v1, v2) => {
   let d = v2.x - v1.x
   let n = v2.y - v1.y
   return d === 0 ? undefined : n / d
+}
+
+// computes a bounding box from a set of points.
+export const findBounds = (vertices) => {
+  let n = vertices.length
+
+  if (n === 0) {
+    return []
+  }
+
+  let loX = vertices[0].x
+  let hiX = vertices[0].x
+  let loY = vertices[0].y
+  let hiY = vertices[0].y
+
+  for(let i=1; i<n; ++i) {
+    let p = vertices[i]
+    loX = Math.min(loX, p.x)
+    hiX = Math.max(hiX, p.x)
+    loY = Math.min(loY, p.y)
+    hiY = Math.max(hiY, p.y)
+  }
+
+  return [new Victor(loX, loY), new Victor(hiX, hiY)]
+}
+
+// resizes each vertex in a list to the specified dimensions. Will not stretch the shape.
+export const resizeVertices = (vertices, sizeX, sizeY) => {
+  let size = Math.max(sizeX, sizeY)
+  let bounds = findBounds(vertices)
+  let curveSize = Math.max(bounds[1].y - bounds[0].y, bounds[1].x - bounds[0].x)
+  let scale = size/curveSize
+
+  let scaledBounds = [bounds[0].multiply({x: scale, y: scale}), bounds[1].multiply({x: scale, y: scale})]
+  let deltaX = size - sizeX - scaledBounds[1].x - scaledBounds[0].x - 1
+  let deltaY = size - sizeY - scaledBounds[1].y - scaledBounds[0].y - 1
+
+  return vertices.map(vertex => vertex.multiply({x: scale, y: scale}).add({x: deltaX/2, y: deltaY/2}))
 }
--- a/src/shapes/Shape.js
+++ b/src/shapes/Shape.js
@@ -1,7 +1,8 @@
 export const shapeOptions = {
   startingSize: {
     title: 'Starting size',
-    min: 1
+    min: 1,
+    isVisible: (state) => { return state.canChangeSize }
   }
 }
 
@@ -14,8 +15,9 @@ export default class Shape {
     return {
       repeatEnabled: true,
       canTransform: true,
-      selectGroup: "Shapes",
-      startingSize: 10
+      canChangeSize: true,
+      selectGroup: 'Shapes',
+      startingSize: 10,
     }
   }
 
--- a/src/shapes/space_filler/SpaceFiller.js
+++ b/src/shapes/space_filler/SpaceFiller.js
@@ -55,7 +55,6 @@ export default class SpaceFiller extends Shape {
     }
   }
 
-  // Algorithm adapted from https://editor.p5js.org/simontiger/full/2CrT12N4
   getVertices(state) {
     const machine = state.machine
     const shape = state.shape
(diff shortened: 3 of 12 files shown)


src/common/geometry.js
```diff
@@ -31,3 +33,41 @@ export const slope = (v1, v2) => {
   let n = v2.y - v1.y
   return d === 0 ? undefined : n / d
 }
+
+// computes a bounding box from a set of points.
+export const findBounds = (vertices) => {
```
jeffeb3, line 38: Ah, I just needed this function.

src/common/geometry.js
```diff
@@ -31,3 +33,41 @@ export const slope = (v1, v2) => {
...
+    hiX = Math.max(hiX, p.x)
+    loY = Math.min(loY, p.y)
+    hiY = Math.max(hiY, p.y)
+  }
+
+  return [new Victor(loX, loY), new Victor(hiX, hiY)]
+}
+
+// resizes each vertex in a list to the specified dimensions. Will not stretch the shape.
+export const resizeVertices = (vertices, sizeX, sizeY) => {
```
jeffeb3, line 62: A lot of this looks very similar to the stuff I just finished to stretch the gcode to fill the space.

src/shapes/Shape.js
```diff
@@ -1,7 +1,8 @@
 export const shapeOptions = {
   startingSize: {
     title: 'Starting size',
-    min: 1
+    min: 1,
+    isVisible: (state) => { return state.canChangeSize }
```
jeffeb3, line 5: :+1:

jeffeb3:
I understand that these are built based on the level, recursively building up to be more and more complicated with each level. I then see that you're sort of masking and stretching the results to fill the space exactly.

Since these are space fillers though, does it make sense to define the complexity by some rough estimate of the distance between the pattern lines? Because my machine is a set size (and yours is too, although thetarho patterns don't have units). So if I am always running with a 1/2" steel ball, there is a reasonable choice to fill the space with each pattern. This is the logic behind the wiper's size parameter.

I am just thinking about this "out loud".

To do that, you'd probably need to always make a pattern that was too large, and too complex, and scale and crop it to have the right line segment size.

jeffeb3:
How hard would it be to make it always start or end in a corner/perimeter or center? Do any of these end in the center?

src/shapes/space_filler/SpaceFiller.js
```diff
@@ -0,0 +1,175 @@
...
+      grid,
+      rotation,
+      size
+    } = config
+
+    if (order === 1) {
+      return seedPointsOrdered
+    }
+
+    const prevOrder = this.fillSpace(order - 1, config)
```
jeffeb3, line 137: Should we add some kind of an error/escape/emergency stop in case this ends up computing way too long/many loops? Just something to error, and return a half pattern, instead of crashing the tab? I hit a condition with the flow snake where it ran out of memory.

src/shapes/space_filler/fillers.js
```diff
@@ -0,0 +1,87 @@
+// L-system instructions for space filling curves
+
+export const fillers = {
+  // http://mathforum.org/advanced/robertd/lsys2d.html
+  'Gosper (flowsnake)': {
```
jeffeb3, line 5: These need to match exactly the keys in the option. Should we use these keys() to generate the options?

src/shapes/space_filler/SpaceFiller.js
```diff
@@ -0,0 +1,175 @@
...
+
+    config.seedPointsOrdered = seedPointsOrdered
+    config.size = size
+
+    // recursively generate our curve
+    let curve = this.fillSpace(order, config)
+    return curve.map(vertex => vertex.add({x: -size*config.grid/2, y: -size*config.grid/2}))
+  }
+
+  // Algorithm adapted from https://editor.p5js.org/simontiger/full/2CrT12N4
```
jeffeb3, line 121: Is there any way to contact this simontiger and make sure they are ok with us publishing it with an MIT license? I doubt they will care, but I like to give them the opportunity to say yes.

jeffeb3:
In general. This all looks very awesome. The patterns are fun, and they are definitely eye catching.The math also has a pretty fun back story.

It makes me really wish we had a better machine preview, because I know these patterns will have a much different feel when applied with a marble. Not that I'm interested in working on that. I think it would probably bring it to a crawl.

I have a couple of ideas to maybe make it better, but I would accept this as-is. It is definitely a huge addition.

Thank you!

src/common/geometry.js
bobnik: Should we attempt to reconcile afterward, assuming this gets merged?

bobnik:
> does it make sense to define the complexity by some rough estimate of the distance between the pattern lines? Because my machine is a set size (and yours is too, although thetarho patterns don't have units). So if I am always running with a 1/2" steel ball, there is a reasonable choice to fill the space with each pattern. This is the logic behind the wiper's size parameter.
> To do that, you'd probably need to always make a pattern that was too large, and too complex, and scale and crop it to have the right line segment size.

I see what you are saying. I think it comes down to the definition of "eraser". I kind of like "overlay" better, because there are times when you don't want to erase everything - you want to overlay a pattern on top or under something. As example, here is an image of the Sierpinsky curve run twice at different sizes.

image

I also think it's visually interesting to see how the pattern repeats. That said, if you'd rather I come up with a set resolution, let me know. It would be straightforward to modify.

src/shapes/space_filler/fillers.js
bobnik: I didn't because I intend to create another non-eraser shape that uses all kind of L-system patterns and I was going to include them in this list. But I think you are right - I can have 2 different lists. I'll make this change.

src/shapes/space_filler/SpaceFiller.js
bobnik: Flow snake has a high max, because it looks pretty, but it is computationally intensive. I can reduce the max. Regarding a limit, we may want that at a higher level (total loops or whatever), particularly once we allow multiple layers.

src/shapes/space_filler/SpaceFiller.js
bobnik: > How hard would it be to make it always start or end in a corner/perimeter or center? Do any of these end in the center?
I think they all start/end on the perimeter.

bobnik:
> How hard would it be to make it always start or end in a corner/perimeter or center? Do any of these end in the center?

I think they all start/end on the perimeter.

src/shapes/space_filler/SpaceFiller.js
bobnik: We only use his code for the Morton curve. I'll either reach out, find another option, or remove it.

jeffeb3:
> I also think it's visually interesting to see how the pattern repeats. That said, if you'd rather I come up with a set resolution, let me know. It would be straightforward to modify.

I was thinking of it as a parameter, like the wiper and spiral wiper have the distance between paths. It is helpful for me to think that way, because 12mm is the size of the steel ball, so it is easy for me to get from there to a pattern that will clean up the whole board.

But I am also fine with this. This is the best chance to change it. After we get save/load functions, we won't be able to change things like that.

jeffeb3:
I merged this. I think this is great. But nothing is set in stone, we can revisit these things if we want to in the future.

PR #129, opened 2020-04-19
Added a simple spiral wiper.

### LINKED ISSUE #111: Circular Eraser
I think it makes sense to have a circular eraser. I am worried that right after that, I'm going to want a polygon eraser... But for now, I think it makes sense to just have something similar to a wiper, but creates a circular pattern. 

Original issue: #65

jeffeb3:
Closes #111 

There will probably be a small conflict with the space filler PR, since we are both editing the same line in the registeredShapes. I'm fine waiting on this until that PR gets merged.

I also added the canChangeSize: false to this shape options, even though it won't do anything until that PR is merged.

screenshot-2020-04-19-1587315956


bobnik:
Would you consider adding this into the existing Wiper as a type, instead of its own shape? So, something like, wiper type = ['Line', 'Circle']. I could see adding other simple types like 'Zig-zag'.

jeffeb3:
> Would you consider adding this into the existing Wiper as a type, instead of its own shape? So, something like, wiper type = ['Line', 'Circle']. I could see adding other simple types like 'Zig-zag'.

I could see that. I thought this was going to be more complicated, so I wanted a "clean slate" to look at, but it really isn't complicated and can live in either spot. I was also trying to give the other erasers some company in the list.

jeffeb3:
Along the lines of a zig-zag, I could also see some simple waves being pretty neat. 

The (Linear) wiper code is pretty long, really. Especially now that we have the perimeter optimizations done, I bet we can reduce the complexity a lot by just having a large grate and let the limits code clip it down. At any rate, it probably makes sense to move this stuff into it's own folder and split it up a bit if we make more wipers.

bobnik:
One of my all time favorite erase patterns is default Sisyphus one called 'CWarp3B'. Here it is with another custom (and epic) pattern on top, but see the background pattern. It's a spiral that oscillates around the spiral line. Could we add an option for something like that?

image

This obviously doesn't need to go into this PR - just musing. I am wondering if introducing a random "noise" element could be cool, too - you could allow the user to set the seed if we felt it needed to be deterministic.

bobnik:
PR looks great/ready to merge.

Issue #130, opened 2020-04-20
Web Hypocycloid


khgrap:
Under the Web shape tool, it seems as if there's a limit to how many rotations are allowed before the pattern re-starts.  This works out to some nice symmetric and pleasing patterns if the "large circle radius" is perfectly divisible by the "small circle radius."  It would be nice if there was an option to allow the pattern to continue without it "jumping" back to the same (relative) starting point. For example, with Large set to 5 and Small set to 3, after one rotation, the chain "jumps" to the beginning of the next loop in line with the first point of the first loop.  If the pattern were to be allowed to continue, the points would converge after 3 rotations (with no growth). The linked Wolfram Hypocycloid page shows this.  Is this possible?

bobnik:
It's not possible now, but it can easily be supported. I experimented with modifications that would allow this shape (and its related shape, "Clover") to continue to rotate until they complete their pattern as you suggest. It radically changes the way these shapes draw as you change the values, but I think it's an improvement. I'll submit a PR that explores this.

Here is large circle = 190, small circle = 37 (rotates 37 times to complete the pattern).
![image](https://user-images.githubusercontent.com/3259145/79747922-7ecbfa80-82da-11ea-9dca-ea192ac4c900.png)

khgrap:
Neat pattern.  This is great - thanks for exploring!

jeffeb3:
@khgrap, This update has been merged, but won't be deployed until the next release.

khgrap:
Sounds great. Thanks!

PR #131, opened 2020-04-20
rotate until the "pattern" is complete

bobnik:
This PR is a response to #130. To produce more pleasing art, when _a/b_ is a non-integer, will rotate _b_ times. I also made the two variables whole numbers instead of decimals. That makes it easier to replicate the examples on the Wolfram Alpha pages.

The shapes now produce quite different results than before with some values. My opinion is that this is an improvement and that shapes that need extra rotations look odd when they cannot complete. That said, it _is_ a change and I'm not sure if we need to worry about this?

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,8 @@
   "dependencies": {
     "@reduxjs/toolkit": "^1.2.5",
     "color": "^3.1.2",
+    "core-js": "^3.6.5",
+    "jest-canvas-mock": "^2.2.0",
     "lodash": "^4.17.15",
     "node-sass": "^4.13.1",
     "rc-slider": "^8.6.3",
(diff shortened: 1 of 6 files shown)


jeffeb3:
Doesn't it often repeat before b? Should we try b/2 and b/3 or something?

I was going to suggest this same change though. I think it is good.

bobnik:
> Doesn't it often repeat before b? Should we try b/2 and b/3 or something?

Yes, you are right. I'll push a fix later. I also am trying to get the shape to stay a constant normalized size (so that starting size is the only parameter you have to fiddle with)

bobnik:
The npm test check failed, but it seems to be because it is trying to run the skipped "App.test.js" spec. Not sure why this is happening...

jeffeb3:
Weird. That test came from the original example I started with (create-react-app). I like the idea of that test, but there's no reason to keep it if it doesn't work, and this PR is not breaking it (it is just broken). So if you don't want to fix it, I'm fine removing it.

bobnik:
> Weird. That test came from the original example I started with (create-react-app). I like the idea of that test, but there's no reason to keep it if it doesn't work, and this PR is not breaking it (it is just broken). So if you don't want to fix it, I'm fine removing it.

I went ahead and fixed the spec, and also the react-ga warning we were getting.

PR #133, opened 2020-04-22
new l-system shape; some l-system refactoring

bobnik:
This PR takes the l-system introduced in the Space Filler shape, and repurposes it as a general shape. This gives us more flexibility in the types of fractal patterns we can create. I've added a number of "classic" fractals.

For the future:

- I would like to allow users to do their own L-system programming, but that's more complex and can come downstream of these stock shapes.
- The current l-system does not yet support branching and a few other cool options that would let us build more patterns. In the case of branching, it does not always produce valid (single-line) patterns when it branches and we'll need to modify a bit if we want to support.


jeffeb3:
Looks great. Thanks!

PR #134, opened 2020-04-25
Feature/l system branching

bobnik:
This PR adds support for l-system branching, which lets us build things like the penrose tile, and a lot of beautiful fractal trees. It also introduces some compute time complexity which limits the shapes we can support. Down the line, when we support worker-based rendering and/or shape caching, we can push the boundaries a bit more.

### DIFF
--- a/src/common/Graph.js
+++ b/src/common/Graph.js
@@ -1,77 +1,122 @@
 import { PriorityQueue } from './PriorityQueue.js'
+import Victor from 'victor'
 
-export const mix = (u, v, s) => {
+export const mix = (v1, v2, s) => {
+  let u = [v1.x, v1.y]
+  let v = [v2.x, v2.y]
   var result = []
+
   for (var i = 0; i < u.length; ++i ) {
     result.push((1.0 - s) * u[i] + s * v[i])
   }
 
-  return result
+  return new Victor(result[0], result[1])
 }
 
-export const vec2 = (...args) => {
-  switch (args.length) {
-    case 0: args.push(0.0); break
-    default: args.push(0.0)
-  }
-
-  return args.splice(0, 2)
-}
-
+// note: requires string-based nodes to work properly
 export default class Graph {
   constructor() {
-    this.nodes = []
+    this.nodeMap = {}
+    this.nodeKeys = new Set()
     this.adjacencyList = {}
-    this.edges = []
+    this.edgeMap = {}
+    this.edgeKeys = new Set()
+    this.clearCachedPaths()
   }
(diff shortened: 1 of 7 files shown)


src/common/Graph.js
```diff
@@ -1,77 +1,122 @@
...
-    case 0: args.push(0.0); break
-    default: args.push(0.0)
-  }
-
-  return args.splice(0, 2)
+  return new Victor(result[0], result[1])
 }
 
+// note: requires string-based nodes to work properly
 export default class Graph {
```
jeffeb3, line 17: I don't want to make things too complicated, and make work for the sake of work. But would it be interesting to try to make some of this Graph stuff into it's own npm library? I think it would add some complication, especially when trying to update it, but it might be something to think about.

bobnik: I have been looking at changing over to this library: https://github.com/trekhleb/javascript-algorithms. I keep having to find and add various graph algorithms, and he's written them all. As you said, it's more work though... I have my sights set on a new shape which needs may push me in this direction :)

jeffeb3:
I really like playing with these, and I can tell when it is reading from the cache. Some things definitely take some time though, like changing the starting size. 

I can't shake the feeling that there is a better way to handle backtracking than dijkstra. Things like the fractal tree should really not take that long to solve. I am more than happy to merge this as-is, and I don't know when I would find the time to solve them analytically. I play around with it if I really get the itch.

bobnik:
> I can't shake the feeling that there is a better way to handle backtracking than dijkstra. Things like the fractal tree should really not take that long to solve. I am more than happy to merge this as-is, and I don't know when I would find the time to solve them analytically. I play around with it if I really get the itch.
I agree with you. I couldn't figure it out, and dijkstra's was the "lazy" approach. I struggled for awhile to get this working even then.

src/shapes/lsystem/subtypes.js
```diff
@@ -24,6 +24,51 @@ export const subtypes = {
     minIterations: 1,
     maxIterations: 8
   },
+  // https://onlinemathtools.com/l-system-generator
+  'Fractal Tree': {
+    axiom: 'F',
+    draw: ['F'],
+    rules:  {
+      F: 'F=F[+FF][-FF]F[-F][+F]F',
```
jeffeb3, line 32: Should this have the `F=`?

bobnik: No, and the tree looks way better when it's fixed. Do you want me to submit a PR?

PR #140, opened 2020-04-26
Feature/import export

bobnik:
This PR reorganizes import and export features to make it easier to add support for new formats without having to duplicate a lot of code. In the short term, this will likely be a new .snd (Sandify) format. It also moves us toward an import becoming a new shape. I did not add this because it doesn't work well until we support multiple shapes.

- breaks GCodeGenerator into:
  - Downloader
  - ThetaRhoExporter
  - GCodeExporter
- breaks PatternImport into:
  - Uploader
  - ThetaRhoImporter
  - GCodeImporter
- creates new "models" for Machine, Importer, and Exporter so that field option titles can be re-used in both UI and in export.
- adds new "update" dispatcher actions on each of above slices so that we can remove boilerplate.
- extracts common logic for adding indented lines to a file into Importer class. Refactors getComments and both file exporters to use the new methods.


src/features/exporter/GCodeExporter.js
```diff
@@ -0,0 +1,34 @@
...
+
+    centeredVertices.map(this.gcode).forEach(line => this.line(line))
+  }
+
+  gcode(vertex) {
+    let command = 'G01' +
+      ' X' + vertex.x.toFixed(3) +
+      ' Y' + vertex.y.toFixed(3)
+
+    if (vertex.speed > 0.0) {
```
jeffeb3, line 28: These don't actually have speeds anymore, do they?

src/features/importer/Uploader.js
```diff
@@ -0,0 +1,150 @@
...
+            Name: {this.props.fileName} <br />
+            Number of points: {this.props.vertices.length } <br />
+            Comments:
+            <div className="ml-3">
+              { commentsRender }
+            </div>
+          </div>}
+        </Card>
+
+        <div className="p-4">
```
jeffeb3, line 127: This is clearly not an "InputOption" or any other kind of option. I would still like to move this stuff into the "draw" tab like we talked about, but we want some of this information to go with it (and some to end up in the about tab).

Does it make sense to have some other kind of "option" that can just hold stuff like this? I'm not sure how the commentsReader stuff will work either. I know I could make something like a `CheckboxOption` that would just be for drawing comments. Is that the right way to go? Or should it be some kind of `StaticOption` that just contains whatever html we want?

jeffeb3: I feel like we need to accept more than HTML, since these are all react objects.

src/features/importer/importerSlice.js
```diff
@@ -0,0 +1,28 @@
+import { createSlice } from '@reduxjs/toolkit'
+
+const importerSlice = createSlice({
+  name: 'importer',
+  initialState: {
+    name: '',
```
jeffeb3, line 6: Is this supposed to be `name` or `fileName`?

jeffeb3: I think fileName. I found two other places. I will push a fix.

src/features/machine/RectSettings.js
```diff
@@ -26,7 +28,8 @@ const mapStateToProps = (state, ownProps) => {
     minY:    state.machine.minY,
     maxY:    state.machine.maxY,
     origin:   state.machine.rectOrigin,
-    minimizeMoves: state.machine.minimizeMoves
+    minimizeMoves: state.machine.minimizeMoves,
+    options: new Machine().getOptions()
```
jeffeb3, line 32: Doesn't this include a bunch of the other fields above?

src/models/Transform.js
```diff
@@ -10,20 +10,32 @@ const transformOptions = {
     min: 1
   },
   transformMethod: {
-    title: 'Transform after each',
+    title: 'When transforming shape',
     type: 'dropdown',
     choices: ['smear', 'intact'],
   },
+  growEnabled: {
```
jeffeb3, line 17: I don't understand why these are new. What am I missing?

jeffeb3:
This PR makes a lot of sense to me. The names are much better (importer/uploader). The dropdown in the output is great. I think making a new export format for "angle/radius gcode" would be a lot easier now. I hope it will also be easier to pull the import into the drop down in draw.

This is great. I do have a few questions, but I think it is almost there.

src/features/exporter/GCodeExporter.js
bobnik: You're right. I only ported this code as-is.

src/models/Transform.js
bobnik: The fields are custom toggle-buttons and we didn't need an option for them to render. Now we do, because the CommentsExporter relies on options for its labels. This lets us not have to hardwire labels in both options and in getComments.

bobnik: The fields are custom toggle-buttons and we didn't need an option for them to render. Now we do, because the CommentsExporter relies on options for its labels. This lets us not have to hardwire labels in both options and in getComments.

src/features/importer/Uploader.js
bobnik: I agree we'll need to deal with this, but I think it can wait until import adds a new shape to the playlist.

PR #141, opened 2020-04-27
Analytic path reverse

jeffeb3:
This seems to work well enough in my mind. I didn't make it very clean (I'm outta time, ATM).  I also didn't test this vertex by vertex to the old method, so it may not be the same. But the patterns (fractal tree 1/2, and penrose) look the same in the preview.

I am pretty sure it isn't reversing an enormous amount of the path, because I am only keeping one reversePath open at a time.


bobnik:
> I am pretty sure it isn't reversing an enormous amount of the path, because I am only keeping one reversePath open at a time.
This is fantastic. Less code and much faster. I'm going to up the limits and add a few more shapes to this PR if you don't mind (ones I had to remove due to poor performance!). Or I can add to a new PR...

jeffeb3:
You can add it to this PR. I have a few minutes, I am going to clean up the code in common/lindenmayer.

bobnik:
> I am pretty sure it isn't reversing an enormous amount of the path, because I am only keeping one reversePath open at a time.

I've been playing with this more, and at higher iterations, and its adds a 3X multiple to distance.

Penrose (5 iterations)
Dijsktra's: 98K
Reversing path: 283K

Penrose (6 iterations)
Dijsktra's: 270K
Reversing path: 778K

I like the way to more complex patterns look, but they will take forever to draw on the table. Thoughts?

jeffeb3:
> > 
> 
> I've been playing with this more, and at higher iterations, and its adds a 3X multiple to distance.

Is that just with the penrose? I was focusing more on the fractals, but the penrose has those loops (it's a bit hard to wrap my head around). I am not sure which way dijkstra is finding around the loop.

> I like the way to more complex patterns look, but they will take forever to draw on the table. Thoughts?

IDK. I like a snappy UI, but I also spend more time in developing sandify than I do watching patterns play out. Maybe I should play the pattern each way and see which one looks better.

PR #143, opened 2020-04-29
fix weirdness with perimeter detection

bobnik:
This PR fixes a few edge cases where perimeter optimization does not work as expected. It has the pleasant side effect of "maxing out" the perimeter distance if you start to draw shape lines entirely out of bounds. I don't think limiting vertex precision to .001 will affect other shapes, but let me know.

### DIFF
--- a/src/features/machine/PolarMachine.js
+++ b/src/features/machine/PolarMachine.js
@@ -88,13 +88,13 @@ export default class PolarMachine extends Machine {
   }
 
   // Returns whether a given path lies on the perimeter of the circle.
-  onPerimeter(v1, v2, delta=.001) {
+  onPerimeter(v1, v2, delta=1) {
     let rm = Math.pow(this.settings.maxRadius, 2)
     let r1 = Math.pow(v1.x, 2) + Math.pow(v1.y, 2)
     let r2 = Math.pow(v2.x, 2) + Math.pow(v2.y, 2)
     let d = this.perimeterDistance(v1, v2)
 
-    return (r1 >= rm - delta && r2 >= rm - delta) && d < 15
+    return Math.abs(r1 - rm) < delta && Math.abs(r2 - rm) < delta && d < 4*Math.PI
   }
 
   // The guts of logic for this limits enforcer. It will take a single line (defined by
--- a/src/features/machine/PolarMachine.js
+++ b/src/features/machine/PolarMachine.js
@@ -94,7 +94,14 @@ export default class PolarMachine extends Machine {
     let r2 = Math.pow(v2.x, 2) + Math.pow(v2.y, 2)
     let d = this.perimeterDistance(v1, v2)
 
-    return Math.abs(r1 - rm) < delta && Math.abs(r2 - rm) < delta && d < 4*Math.PI
+    // Delta is purposefully large to accommodate the squaring of the compared values.
+    // Setting delta too small will result in perimeter moves being miscategorized.
+    // d is used to guard against the case where there is a straight line connecting two
+    // perimeter points directly. In this case, we want to register that as a non-perimeter
+    // move, or it will be incorrectly optimized out of the final vertices. The 3/50
+    // ratio could likely be refined further (relative to maxRadius), but it seems to produce
+    // accurate results at various machine sizes.
+    return Math.abs(r1 - rm) < delta && Math.abs(r2 - rm) < delta && d < 3*this.settings.maxRadius/50
   }
 
   // The guts of logic for this limits enforcer. It will take a single line (defined by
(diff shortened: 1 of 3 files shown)


src/features/machine/PolarMachine.js
```diff
@@ -88,13 +88,13 @@ export default class PolarMachine extends Machine {
   }
 
   // Returns whether a given path lies on the perimeter of the circle.
-  onPerimeter(v1, v2, delta=.001) {
+  onPerimeter(v1, v2, delta=1) {
```
jeffeb3, line 91: This 1 is a much bigger delta. Does it need to just be bigger than the precision? Maybe 0.01 or 0.005 would be better?

src/features/machine/PolarMachine.js
```diff
@@ -88,13 +88,13 @@ export default class PolarMachine extends Machine {
...
     let rm = Math.pow(this.settings.maxRadius, 2)
     let r1 = Math.pow(v1.x, 2) + Math.pow(v1.y, 2)
     let r2 = Math.pow(v2.x, 2) + Math.pow(v2.y, 2)
     let d = this.perimeterDistance(v1, v2)
 
-    return (r1 >= rm - delta && r2 >= rm - delta) && d < 15
+    return Math.abs(r1 - rm) < delta && Math.abs(r2 - rm) < delta && d < 4*Math.PI
```
jeffeb3, line 97: What is the case where d would be greater than 4pi? I'm not sure why there is even a permiterDistance involved in this function.

jeffeb3: Oh, wait, this is trying to determine if the vector from v1 to v2 is along the perimeter (already, not assuming it's following the arc).

So both points need to be near the max radius, and the distance between them needs to be small enough to be approximated by a line or an arc. So -maxRadius,0 to maxRadius,0 would be too far to be approximated by a line.

4pi is pretty close to 15, so I guess it didn't change much. I feel like there should be a /radius in there somewhere.

bobnik:
@jeffeb3, I'll look at the polar onPerimeter later today - I agree the math doesn't add up logically, though it appears to work for the cases I tried. Also, to see the fixes in action, try your gcode import that gave weird perimeter lines. Also, try Hilbert Space Filler on rectangular machine at level 5 (before/after).

src/features/machine/RectMachine.js
```diff
@@ -61,15 +61,11 @@ export default class RectMachine extends Machine {
...
+  onPerimeter(v1, v2, delta=.0001) {
     const dx = Math.abs(Math.abs(v1.x) - this.sizeX)
     const dy = Math.abs(Math.abs(v1.y) - this.sizeY)
 
-    if ((v1.x === v2.x && dx < delta) || (v1.y === v2.y && dy < delta)) {
-      return v1.x === v2.x || v1.y === v2.y
-    } else {
-      return false
-    }
+    return (v1.x === v2.x && dx < delta) || (v1.y === v2.y && dy < delta)
```
jeffeb3, line 68:
```suggestion
    return (Math.abs(v1.x - v2.x) < delta && dx < delta) || (Math.abs(v1.y - v2.y) < delta && dy < delta)
```

src/features/machine/PolarMachine.js
jeffeb3: If you are trying to limit it to a specific angle (in radians) it would be:

d / (2pi*r) < (angle/2pi)
d < angle * maxRadius

A specific arc length would also make some sense. 

If you wanted to determine the max distance between the arc and the connecting line, it would happen halfway between the two points, and it would be the difference between the long side and the maxRadius.

jeffeb3: That looks good, but it is totally garbage.. Not sure what the problem is. Maybe the small angle approx. is removing too much of the error. I have to put in pretty crazy numbers to get 15 out of it.

jeffeb3: Without the angle approx:
error = maxRadius(1-cos(d/2maxRadius))
d < acos(1-maxError/maxRadius)*2maxRadius

If maxError is 0.1 and maxRadius is 400, that comes up with a max d of 17.9, so that is reasonable.

FTR, I don't really care this much :). I think it is splitting hairs for drawing in sand. It is an interestin problem to think about though.

bobnik: Ok. I tried to document my rationale for the logic used. The case to test, if you are interested, is a Polygon, 4 sides, with a default spin (polar machine). If you don't keep the large delta, as you grow the shape, you'll continue to add distance (it never stabilizes). If you don't add the max. distance check, you'll drop valid polygon lines near the perimeter. Hopefully my inline comments make sense.

Issue #139, opened 2020-04-25
Support shape caching


bobnik:
Calculation of vertices for a base shape, given a set of specific field values, is deterministic. We recalculate the same values every time any change is made. For simple shapes, the performance impact is negligible. For more complex shapes, like some l-systems, it can bog down the UI.

I propose we do some simple in-memory caching of shapes as they are rendered. So, you would take a calculation "hit" the first time you drew a shape, but after that, the cached version could be used. We'll have to make sure we expire shapes in the cache as needed to avoid running out of memory - I am not sure what this calculation looks like, but I am hoping we can set some sensible defaults.

jeffeb3:
The wiper doesn't need caching. Some shapes, like the star, have some options that are pretty variable, and they don't benefit from caching.

PR #144, opened 2020-04-29
Feature/cache vertices

bobnik:
Per earlier discussion, this PR works to tweak the state tree into a structure that follows how vertices are generated, and then create nested cached selectors to minimize when vertices are redrawn. From an end-user standpoint, it greatly improves UI responsiveness, particularly with the preview slider and making transformation or machine changes.


src/features/machine/selectors.js
```diff
@@ -1,47 +1,98 @@
...
 const getImporter = state => state.importer
-const getExporter = state => state.exporter
 const getMachine = state => state.machine
 
+// by returning null for shapes which can change size, this selector will ensure
+// transformed vertices are not redrawn when machine settings change
+const getTransformMachine = state => getTransform(state).canChangeSize ? null : state.machine
+
+// requires only shape, and in the case of erasers, machine state
+export const getShapedVertices = createSelector(
```
jeffeb3, line 23: These selectors are the key then? If you cascade them with just the minimum dependencies, it doesn't require all the calculations to happen again? And ciritically, the getCurrentShape is only dependent on the shape items, not the lower down selectors?

src/features/machine/computer.js
```diff
@@ -182,29 +160,35 @@ function reportTiming(time) {
...
-        for (var j=0; j<input.length; j++) {
-          let amount = state.transform.transformMethod === 'smear' ? i + j/input.length : i
-          outputVertices.push(transform(state.transform, input[j], amount, amount))
-        }
+    }
+  } else {
+    for (i=0; i<numLoops; i++) {
+      if (transform.transformMethod === 'smear' && transform.repeatEnabled) {
+        // remove last vertex; we don't want to return to our starting point when completing the shape
+        outputVertices.pop()
```
jeffeb3, line 179: This still isn't quite right.

jeffeb3: I think I fixed it. The startingSize wasn't being applied the right way to the tracks, and the non-smeared shapes weren't being connected the right way. There was also a (probably very old) but where I was iterating over i in a loop, then iterating over i again later, but it was somehow missed, because it was iterating over the same thing.

src/features/machine/selectors.js
```diff
@@ -1,47 +1,98 @@
...
+      getCurrentShape,
+      getTransformMachine
+  ],
+  (shape, machine) => {
+    const state = {
+      shape: shape,
+      machine: machine
+    }
+    const metashape = getShape(shape)
+    return metashape.getVertices(state)
```
jeffeb3, line 34: This seems like the place to store the cache for the metashape base shapes. Storing a list of state.shapes to caches of metashape.getVertices(state), and then recalling them if they are in the cache.

I'm just not sure how to cache them in redux. Do we make another selector for the cache? Is that counter productive?

jeffeb3: This is related to #139

src/features/machine/selectors.js
bobnik: Yes, exactly.

PR #145, opened 2020-05-02
Trying out caching of shape vertices.

jeffeb3:
#139 

This seems to work. I put in some console.log statements and I can tell it is returning the cache. It is also much faster on some of the fractal line patterns.

I think we can safely bump some of those up as much as it makes sense. If you want @bobnik, maybe you could take a look at this branch, and while you are at it, adjust the max for those lines to be where you want them? Even if this completely makes the performance fine (which it probably doesn't). I think there is still probably a reasonable max for the patterns.

I also turned off the caching of the space filler l system. I figured that since it uses the machine limits, the caching wouldn't really make sense, and it would probably be an enormous cache to cache every machine size.

### DIFF
--- a/src/models/Shape.js
+++ b/src/models/Shape.js
@@ -26,18 +26,6 @@ export default class Shape {
     return shapeOptions
   }
 
-  getVerticesWithCache(state) {
-    this.cache.forEach( (cachedShape) => {
-      var [shape, vertices] = cachedShape
-      if (JSON.stringify(state.shape) === JSON.stringify(shape)) {
-        return vertices
-      }
-    })
-    const shapeVertices = this.getVertices(state)
-    this.cache.push([state.shape, shapeVertices])
-    return shapeVertices
-  }
-
   getVertices(state) {
     return []
   }
--- a/src/models/Shape.js
+++ b/src/models/Shape.js
@@ -3,6 +3,7 @@ export const shapeOptions = {}
 export default class Shape {
   constructor(name) {
     this.name = name
+    this.cache = []
   }
 
   getInitialState() {
@@ -10,6 +11,7 @@ export default class Shape {
       repeatEnabled: true,
       canTransform: true,
       selectGroup: 'Shapes',
+      shouldCache: true,
     }
   }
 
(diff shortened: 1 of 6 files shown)


src/models/Shape.js
```diff
@@ -24,6 +26,18 @@ export default class Shape {
     return shapeOptions
   }
 
+  getVerticesWithCache(state) {
+    this.cache.forEach( (cachedShape) => {
+      var [shape, vertices] = cachedShape
+      if (JSON.stringify(state.shape) === JSON.stringify(shape)) {
```
jeffeb3, line 32: The Internet said this was fragile, but the other suggestions looked terrible. Is this the right way to compare the state.shape objects?

jeffeb3:
I tried to play with this for a few minutes. The memory snapshots I can take in firefox show it at 20MB when it starts, and after going through a bunch of shapes, including all the fractal lines and the existing options, I got it up to 165MB. The memory showed a lot of that in Arrays and Objects, which makes sense.

I don't have a good idea if that is a lot or not. It didn't take long to get twitter up to 150MB. Slack starts at 120MB. So maybe it's not so bad, since that might be close to an upper limit. Maybe I'll just keep try to remember to keep an eye on it when we add any new features. I think we could probably get away with only keeping the last 20-30 of the most recent caches, but the easy way I did this has a different cache for each Shape.

jeffeb3:
It might be smaller when it is deployed to. Not sure how to test that locally.

bobnik:
> I also turned off the caching of the space filler l system. I figured that since it uses the machine limits, the caching wouldn't really make sense, and it would probably be an enormous cache to cache every machine size.

I pushed a commit with some minor adjustments. We can back out as needed.

- use a LRU cache to manage cache size. I set a limit of 500000 vertices. This is somewhat arbitrary, but I played around for awhile without filling the cache. You can uncomment the logging to see it in action and/or bump up the limit. I think even when fills up, the user experience is good.
- I originally thought it would be ok to cache space fillers, mainly because you generally switch between them and it would be handy to cache the expensive ones (and you don't generally change machine settings that much). That said, what we _need_ to do is pull machine state out of the shape generation, and resize if needed afterward to fill the available machine space. So, please revert this if you think it's detrimental.

jeffeb3:
Thanks. This is great.

Issue #148, opened 2020-05-02
SVG output


jeffeb3:
@msurguy suggested that an SVG output would be useful. Given the new organization of exporters and downloaders, that should be pretty easy.

Are there particular consumers of svg files we should be targetting? Are there drawing bots that use SVG inputs directly? What limitations do those output files have? This is an area I am not too familiar with.

jeffeb3:
SVG Output is on master (but not yet deployed). Here is a link to a file, if someone wants to confirm it actually works with their drawing machines. This one is 500x500 sized.

https://drive.google.com/file/d/1dNOClyr6fugviFJvlmF0agehfNpYP5-E/view?usp=sharing

PR #149, opened 2020-05-07
Feature/svg export

jeffeb3:
# Description
Addressing #148.

This outputs the vertices in SVG format. All the vertices are exported as a single path, with no curves, just lineTo commands in d3.

# Verification

I can look at the file with my text editor, my image viewer, firefox, and inkscape. It looks reasonable for the cases I've tested.

I do not have an axidraw machine though, so I could use some help making sure it actually works on the intended platform.

### DIFF
--- a/src/features/exporter/Downloader.js
+++ b/src/features/exporter/Downloader.js
@@ -14,7 +14,12 @@ import ReactGA from 'react-ga'
 import ThetaRhoExporter from './ThetaRhoExporter'
 import GCodeExporter from './GCodeExporter'
 import SvgExporter from './SvgExporter'
-import Exporter from '../../models/Exporter'
+import {
+  gcodeTypeName,
+  thrTypeName,
+  svgTypeName,
+  Exporter,
+} from '../../models/Exporter'
 
 const mapStateToProps = (state, ownProps) => {
   return {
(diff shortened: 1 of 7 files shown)


src/features/exporter/Downloader.js
```diff
@@ -96,7 +100,11 @@ class Downloader extends Component {
     let link = document.createElement('a')
     link.download = fileName
 
-    let blob = new Blob([text],{type: 'text/plain;charset=utf-8'})
+    let fileType = 'text/plain;charset=utf-8'
+    if (this.props.fileType === 'SVG (.svg)') {
```
jeffeb3, line 104: I don't like using this key string over and over. I would also like to easily change it, since it is user facing.

src/features/exporter/Downloader.js
```diff
@@ -25,18 +26,21 @@ const mapStateToProps = (state, ownProps) => {
     shape: state.shapes.currentId,
     offsetX: (state.machine.minX + state.machine.maxX) / 2.0,
     offsetY: (state.machine.minY + state.machine.maxY) / 2.0,
+    width:   (state.machine.rectangular ? (state.machine.maxX - state.machine.minX) : (2.0 * state.machine.maxRadius)),
```
jeffeb3, line 29: I've been a bit lazy, because most users use rectangular limits with gcode and polar limits with thr, but there's no specific requirement for that. Either it needs to be explicitly coupled somehow, or support criss-cross of these.

src/models/Exporter.js
```diff
@@ -6,22 +11,24 @@ const exporterOptions = {
...
+    type: 'textarea',
+    isVisible: (state) => { return state.fileType !== svgTypeName },
   },
   pre: {
     title: 'Program start code',
-    type: 'textarea'
+    type: 'textarea',
+    isVisible: (state) => { return state.fileType !== svgTypeName },
   },
   reverse: {
```
bobnik, line 26: Should reverse be visible when exporting SVG? I am assuming not.

jeffeb3: I think it could be useful. Besides making svgs for image viewing, I am intending these to be used on axidraw type machines. I assume they don't care much about the direction, but they might care some? IDK.

src/features/exporter/Exporter.js
```diff
@@ -5,37 +5,55 @@ export default class Exporter {
...
     this.exportCode(vertices)
     this.line()
 
-    this.startComments()
-    this.line('BEGIN POST')
-    this.endComments()
-    this.line(this.props.post, this.props.post !== '')
-    this.startComments()
-    this.line('END POST')
+    if (this.props.pose !== '') {
```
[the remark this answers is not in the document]
jeffeb3: And then, in my day job, I'm writing 'post' instead of 'pose' all over the place. :)

Issue #117, opened 2020-04-07
Rad Lines Tool


jeffeb3:
Original issue text from msurguy:

I would love to see if my Rad Lines tool's SVG output could be integrated into this software... 
Here's the tool: https://msurguy.github.io/rad-lines/ 

Here are some examples of the output: 

<img width="338" alt="Screen Shot 2020-04-06 at 10 10 49 PM" src="https://user-images.githubusercontent.com/585833/78632356-a4292500-7853-11ea-90ad-f20a4a8d8611.png">

<img width="528" alt="Screen Shot 2020-04-06 at 10 11 38 PM" src="https://user-images.githubusercontent.com/585833/78632347-a2f7f800-7853-11ea-9efd-afc4eba14799.png">
<img width="592" alt="Screen Shot 2020-04-06 at 10 11 30 PM" src="https://user-images.githubusercontent.com/585833/78632354-a3908e80-7853-11ea-9d54-ae9a05f2fa83.png">

<img width="562" alt="Screen Shot 2020-04-06 at 10 11 46 PM" src="https://user-images.githubusercontent.com/585833/78632342-a0959e00-7853-11ea-8c74-c7eaa1527c8a.png">

_Originally posted by @msurguy in https://github.com/jeffeb3/sandify/issues/110#issuecomment-610177647_

jeffeb3:
Those are some great patterns. The website also looks great, and was really nice to use. I especially like the input equations, they can really expand the capabilities for more advanced users.

I think getting those patterns into sandify would be really great. I don't know yet the best way to do it. If I just had an input for .svgs, I think that would get pretty complicated, since the format supports so much. Exporting the vertices from Rad Lines would work, something in .json format or something.

The bigger issue is that the shapes are not connected, so they wouldn't work in sandify, which never lifts the pen. The original intention is to make patterns from one continuous line, for sand tables that can't lift the marble.

I could see incorporating some of the features from Rad Lines into the shapes and transforms here. Having an editable equation for the spin and grow would be pretty neat. I already have an issue to round over the corners of the polygon. WDYT? I don't want to step on anyone's toes.

msurguy:
@jeffeb3 couple ideas here to address your points:

- after the individual curves are generated, we can connect the start point of each curve to generate single path
- the gist of the code that is making the lines is actually pretty simple, https://github.com/msurguy/rad-lines/blob/master/src/components/Polygons.vue#L218
[...]

PR #151, opened 2020-05-08
Rounded corners on polygons.

jeffeb3:
Creates round corners (part of #117).

Since we don't handle arcs any differently, and we smear evenly based on the vertex index, leaving the line segments as lines means they get skewed funny compared to the arcs. Ideally, the arc length is basically the same as the line length, which I think I have gotten close to. But the result is a 4 vertex shape just became a 200 vertex shape. Not a huge deal, but it's frustrating we can't smear an arc more smoothly.

### DIFF
--- a/src/models/Polygon.js
+++ b/src/models/Polygon.js
@@ -7,7 +7,18 @@ const options = {
     polygonSides: {
       title: 'Number of sides',
       min: 3
-    }
+    },
+    roundCorners: {
+      title: 'Round Corners',
+      type: 'checkbox',
+    },
+    roundFraction: {
+      title: 'Round Fraction',
+      min: 0.0,
+      max: 1.0,
+      step: 0.05,
+      isVisible: (state) => { return state.roundCorners }
+    },
   }
 }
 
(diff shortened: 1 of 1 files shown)


jeffeb3:
screenshot-2020-05-08-1588978108

Looks pretty neat.

src/models/Polygon.js
```diff
@@ -7,7 +7,18 @@ const options = {
     polygonSides: {
       title: 'Number of sides',
       min: 3
-    }
+    },
+    roundCorners: {
+      title: 'Round Corners',
```
bobnik, line 12: For consistency, I would not capitalize more than the first word in the labels... 'Round corners' and 'Round fraction'

src/models/Polygon.js
```diff
@@ -7,7 +7,18 @@ const options = {
...
+      type: 'checkbox',
+    },
+    roundFraction: {
+      title: 'Round Fraction',
+      min: 0.0,
+      max: 1.0,
```
bobnik, line 18: With a min of 0, you get no rounding and with a max of 1.0 you get a circle no matter how many polygon sides. I'm torn on whether it's good to be able to visualize these extremes, vs perhaps going with a min = 0.1, and max = 0.85 to reflect that "round corners" does what it advertises.

src/models/Polygon.js
```diff
@@ -21,16 +32,56 @@ export default class Polygon extends Shape {
       ...super.getInitialState(),
       ...{
         type: 'polygon',
-        polygonSides: 4
+        polygonSides: 4,
+        roundCorners: false,
+        roundFraction: 0.5,
```
bobnik, line 37: Would you consider making this a slightly lower value? I realize it's subjective, but I like 0.25 as the default value.

src/models/Polygon.js
jeffeb3: You're definitely right. I also think more fine grained detail in the lower range is helpful, and the upper range isn't very interesting, so I changed it to 0.05, 0.5, with a step of 0.025. That looks better to me.

jeffeb3:
Thanks for the feedback.

PR #152, opened 2020-05-09
Feature/math transforms

jeffeb3:
This is the other half of #117.

This incorporates mathjs and it's evaluate function to provide users the ability to use math to create functions for spin and grow.

screenshot-2020-05-09-1589059142

I just used mathjs evaluate directly, without customizations.

The grow function is OK, but the spin function is really fun.


jeffeb3:
Having the growValue multiplied by the output of the equation is really nice. I think having something like that for the spin would be good too. It would be easy to use the spinValue, but the names are a little weird (for both, honestly).

bobnik:
> It would be easy to use the spinValue, but the names are a little weird (for both, honestly).
How about something like this (in order)?

Scale amount
Scale by [constant|custom function] - sticky buttons like [smear|keep intact]
Scale function (if function selected)

Spin amount
Spin by [constant|custom function]
Spin function (if function selected)

src/features/machine/computer.js
```diff
@@ -47,20 +48,42 @@ export const transformShape = (data, vertex, amount, trackIndex=0, numLoops) =>
   let transformedVertex = vertex
 
   if (data.repeatEnabled && data.growEnabled) {
-    transformedVertex = scale(transformedVertex, 100.0 + (data.growValue * amount))
+    var growAmount = 100
```
bobnik, line 51: Super nitpicky, but I would try to stop using 'var' in favor of either 'let' or 'constant'

src/features/machine/computer.js
```diff
@@ -47,20 +48,42 @@ export const transformShape = (data, vertex, amount, trackIndex=0, numLoops) =>
...
+    if (data.growAdvanced) {
+      try {
+        growAmount = data.growValue * evaluate(data.growMath, {i: amount})
+      }
+      catch (err) {
+        console.log(err)
```
bobnik, line 57: If you are leaving this for tech-savvy users, you might prefix it with "error parsing grow function" (similar for spin). If you aren't, you might comment these out.

src/models/Transform.js
```diff
@@ -23,17 +23,37 @@ const transformOptions = {
...
-    step: 0.1
+    title: 'Spin (+/-)',
+    step: 0.1,
+  },
+  spinAdvanced: {
+    title: 'Advanced',
+    type: 'checkbox',
+  },
+  spinMath: {
+    title: 'Spin Function (i)',
```
bobnik, line 50: Reminder that, whatever names we settle on, that first word only is capitalized.

jeffeb3:
> Spin amount
> Spin by [constant|custom function]
> Spin function (if function selected)

This is a little more work. Mostly because it used to be a bool, and now it is spinMethod === 'function'. But I like the look of it, and it is more clear in the comments.

There wasn't quite enough room for "custom function" and it was spilling into two rows, so I changed it to "function" and I think it is just as clear.

Thanks for all the suggestions.

Issue #138, opened 2020-04-25
Imported files should be treated as a custom shape


bobnik:
This issue has a dependent requirement of needing to support multiple shapes, #137. Once we can draw multiple shapes, the import process could be handled from the Playlist panel.

- Support a new Shape type called FileImport (or similar) that cannot be transformed. This shape won’t appear in the Shapes drop list. Attributes are Zoom, File (presumably not editable once selected), and Aspect Ratio.
- On the Layers tab, next to “New” button, add an “Import” button.
 - When clicked, should open a File dialog. If user selects a file, it will add a new FileImport layer. They can resize and move it via the preview.
- Move “where to get .thr files” and “Note about copyrights” to “About”
- Remove Import tab.

jeffeb3:
This is spot on. Thank you. 

I do wonder if allowing transforms, but turning them off would be awesome. Using inkscape or estlcam to create a simple path (think a simple closed shape like a nike swoosh). Then importing it, then transforming it. That would have let me do the V1 logo without any changes to the code.

That is a feature I have wanted from the beginning, but it's never been this close. I may just silently enable this, and then work on some good documentation to explain how it would work later.

PR #153, opened 2020-05-10
let user move and resize shapes from the preview window

bobnik:
This PR lets a user interact with the shape in the preview window. Specifically, move, rotate, and scale the shape. This is mostly done, but it's a work in progress and I want feedback.

I initially thought I would be using 'height' and 'width' to constrain the fully rendered shape, but I backed off from that approach, because it's not how Sandify works with startingSize. Instead, a "transformer" is drawn around the initial shape, and that's what gets transformed (e.g., startingSize, x and y offsets).

To test, click anywhere in the preview window. If you drag, you'll move the shape. If you click, you'll toggle handles to transform the shape.


bobnik:
FYI - there is still some weirdness with konva transforms and our own (spin/track).

bobnik:
> FYI - there is still some weirdness with konva transforms and our own (spin/track).
I am not sure I will get able to konva to scale the shape properly when track is enabled. The transform applies scaling and rotation inside the loop, and konva doesn't work that way. For now, I've solved this by not allowing preview scaling (only rotation) if track is enabled.

jeffeb3:
Ok. I will work on this today. Busy weekend.

jeffeb3:
The transforms have kind of built up in layers, and so, they were a bit hard to unravel.

It seems like a shape should start as a kernel shape, from getVertices. Then go through the loops, with scale, rotate, and track stuff happening still in this "normalized" coordinate frame. Then we go through cache. Then apply the rotate, scale, and offset to the normalized vertices (and the track vertices).

The offsets used to be applied before the rotation. I think you can achieve a similar effect with the tracks now, but those patterns will draw a little differently now. The tracks are neat, but they have always seemed harder to use than the other features. Now I am wondering if there is a way to use the mouse, with the track, to make it more useful.

I have been playing with it, and looking a little bit at the code, and it is very impressive. I am still trying to wrap my head around how it works. It looks like Kona is doing the drag and mouse operations on its own, and the preview is clearly a graphic changing around until you let go, then those parameters go back into the state, and the drawing is redrawn using those parameters. Is that right? I was skeptical of doing that, but with the patterns that have a huge number of vertices, it will be much faster. It works so well, I am convinced.

Is it possible to have an onWheel event (or something) do scaling? What about a right click to rotate? I still would like to keep the handles, so it doesn't have to be part of this PR (and I can play with it, if it doesn't interest you). I haven't really seen where the fun stuff is happening.

Should I be able to grab anywhere in the preview and move it (even way off the shape)? It seems like it is doing some kind of collision detection, and if I move it to one side, and click the free space, it doesn't pick it up. If it is trying to grab only one shape, I wonder what will happen when we have multiple shapes. The collision area seems pretty large.

jeffeb3:
It looks like sometimes the rotation resets to zero. Specifically, when I rotate it to something, and then grab the resize corner and pull it through the middle of the box (so, like a negative scale, maybe?) it resets the rotation to zero.

bobnik:
> Is it possible to have an onWheel event (or something) do scaling? What about a right click to rotate? I still would like to keep the handles, so it doesn't have to be part of this PR (and I can play with it, if it doesn't interest you). I haven't really seen where the fun stuff is happening.

I've pushed a commit to support "wheel" scaling. I am not sure how to handle right-click-rotate. Konva draggable also tracks mouse movements, so you'd have to somehow tell Konva not to drag the shape. Also, I map my right-click on my Mac so it functions as Ctrl-click (to bring up menus and such). If we did support this, we may want to do it via <alt>click vs. right click.

bobnik:
> It looks like Kona is doing the drag and mouse operations on its own, and the preview is clearly a graphic changing around until you let go, then those parameters go back into the state, and the drawing is redrawn using those parameters. Is that right?

Yes. When drag starts, I grab a "dragging" version of the vertices, not bound by machine boundaries. If you don't do this, the clipped boundaries remain visible as you drag around. Konva moves the shape around until dragging stops. I then save the changed offsets to the state and revert back to the non-dragging (machine-bound) vertices.

bobnik:
> It looks like sometimes the rotation resets to zero. Specifically, when I rotate it to something, and then grab the resize corner and pull it through the middle of the box (so, like a negative scale, maybe?) it resets the rotation to zero.

Fixed (I believe).

bobnik:
> Should I be able to grab anywhere in the preview and move it (even way off the shape)? It seems like it is doing some kind of collision detection, and if I move it to one side, and click the free space, it doesn't pick it up. If it is trying to grab only one shape, I wonder what will happen when we have multiple shapes. The collision area seems pretty large.

Yes, this is currently buggy. I can probably fix it so that you can drag from anywhere within machine bounds, but you are right about the collision area being large. Konva defines two functions, sceneFunc and hitFunc. You can read about custom shapes (https://konvajs.org/docs/shapes/Custom.html). 

They recommend you don't do any styling within sceneFunc, but that's not yet possible for us because we change styles to draw our gradient path, end points, etc. If we managed to define a single canvas path inside our paint method, I believe Konva would calculate our "collision area" dynamically. Unfortunately, this would also define our "transform" handles around the entire transformed shape. I don't think we want to constrain our final shape form with constant height/width values. I started down this path, but it makes it difficult to create shapes that are bigger than the preview window. I don't know if I am making any sense... 

TLDR, we aren't a perfect fit for Konva because (1) we want to styling our shapes manually, and (2) we want to define our transform bounds (with handles) using startingSize, and let our machine generate a larger shape. My inclination is to fix the dragging so it works (with the correspondingly large collision area), and punt the larger problem of dragging multiple shapes down the road, when we tackle the playlist.

jeffeb3:
> If we managed to define a single canvas path inside our paint method...

What do you think about the transform handles bigger (but still smaller than the shape)? Something like 5x larger would be easier to grab when it is scaled very small.

jeffeb3:
I pushed a change, because onWheel wasn't working well with the polygon, but I may have broken something else. 

When deltaY was 0, I was getting NaN for a startingSize, so I changed it to only change the size if that was non-zero.

When the startingSize was small, the Log(deltaY) wasn't enough to change the size, so with my wheel, it wasn't doing anything, so I also made it at least 1 (if it wasn't zero).

The problem I'm seeing now is that the shape it loads with works fine, but if I change to another shape, it doesn't actually change that number. I suspect the onWheel function is being created once, and it isn't setting the this.props.transform.startingSize on the right transform when I choose a different shape. It seems like the function is capturing the first transform, and not grabbing the current transform. 

FYI, my mouse is returning multiples of 3 for the deltaY, and I saw events with 0, and a deltaX, which seemed weird to me.

bobnik:
> he problem I'm seeing now is that the shape it loads with works fine, but if I change to another shape, it doesn't actually change that number.

You didn't break this. It's because selectedId isn't changing right away. If you are okay with the handles always being visible on the shape, I believe there's an easy fix. I can look at this evening most likely.

jeffeb3:
I liked the handles not showing all the time, but I am ok with it.

bobnik:
> I liked the handles not showing all the time, but I am ok with it.

I think I've fixed your reported issues. You should be able to drag shape from anywhere in preview, the transformer handles are bigger, and mouse wheel scaling works after changing shape selection.

jeffeb3:
Those changes all look good. The imported vertices are not showing up correctly though. I think they might just be incompatible with these transforms and drawing method in their current state.

If you like, I can merge this through, and then take it as a task to convert the import into a shape, instead of a tab, and then figure out how to fix the scaling, rotating, and offsets. I think it is a little strange the way it has grown up, and it no longer fits the way it is.

bobnik:
> The imported vertices are not showing up correctly though. I think they might just be incompatible with these transforms and drawing method in their current state.

I've fixed the scaling issue in the current branch. While testing this, I noticed that theta-rho import handles offset and starting size correctly, but gcode import does not (or maybe it's the export?). I went back to the original gcode import branch to test and it still doesn't work properly. So I think that this issue(?) was not introduced in this PR.

Regarding move this into a shape (#138), I think that can wait until we get layers working. Otherwise, it doesn't really fit well (and the logic to insert it into the drop list is going to be thrown away anyway).

So... I think it's ready to merge...

jeffeb3:
> While testing this, I noticed that theta-rho import handles offset and starting size correctly, but gcode import does not (or maybe it's the export?). I went back to the original gcode import branch to test and it still doesn't work properly. So I think that this issue(?) was not introduced in this PR.

The import will remove any offset and starting size. Because I don't know where it came from, and I didn't really have a way to adjust it. The next thing I will do is to make the import a shape. I am looking forward to removing some of the hacks in the import, if I can, and also getting a konva handle on an imported pattern.

I tried this and the math stuff together for the first time. It work pretty well together. I am really excited to get this out there.

PR #156, opened 2020-05-15
Only update the math when the user is finished

jeffeb3:
This changes the 2 math fields to have two elements in the transform options. One is the "Input" which react uses to own all the state, and update the field as the user is typing, and the other is the one we are used to, but it only gets updated when there is a "blur" (I had to look this up, it is the opposite of focus, when the user leaves the field) or pressing 'Enter'.

This seems to work pretty well. It is tricky to not break the rest of the InputOption boxes when working on this. I am not completely satisfied with the duplication, but I tried to make it simpler, and it just wasn't working as well.


jeffeb3:
I like that option too, but we want the other inputs to be instantaneous, and we quickly lose the knowledge of what input changed. I'm just not sure where we could actually put the throttle to achieve that.

src/components/InputOption.js
```diff
@@ -6,6 +6,25 @@ import {
...
+
+    if (optionType === 'number') {
+      value = value === '' ? '' : parseFloat(value)
+    }
+
+    attrs[key] = value
+    if (option.onChange !== undefined) {
+      attrs = option.onChange(attrs, model)
+    }
+    this.props.onChange(attrs)
```
jeffeb3, line 25: I think we could probably do this without a doubled state by setting a flag on attrs.

src/components/InputOption.js
```diff
@@ -4,8 +4,18 @@ import {
   Form,
   Row
 } from 'react-bootstrap'
+import debounce from 'lodash/debounce'
 
 class InputOption extends Component {
+  constructor(props) {
+    super(props);
+    this.delayedSet = debounce( (value, key, onChange) => {
```
jeffeb3, line 12: This is a much cleaner option. Thanks Bob.

src/components/InputOption.js
```diff
@@ -4,8 +4,18 @@ import {
...
+      let attrs = {}
+      attrs[key] = value
+      onChange(attrs)
+    }, 1500)
```
jeffeb3, line 16: WDYT about this timeout? It is a bit long, but I liked it not computing while I was typing.

bobnik: 👍🏻

PR #158, opened 2020-05-17
Feature/variable pre post

### LINKED ISSUE #154: Variables in the expory Pre and Post
I'd like to add in some simple variables, that we can use in the pre and post for gcode, and theta rho. Specifically, I want to have the start and stop X values available so I can do things like this in the pre:

```
G1 Z10 F400 ; lift the pen
G1 X{startx} Y{starty} F2000 ; Go to start
G1 Z0 F400 ; drop pen
G1 F2000 ; set speed
```

And then in post
```
G1 Z10 F400 ; lift pen
```

That would be pretty useful for making pen plots on a CNC machine.

I can't come up with every variable we might want, but I am thinking, as a good start:

startx: the first vertex.x
starty: the first vertex.y
endx: the last vertex.x
endy: the last vertex.y
minx: the minimum of the vertices.x
miny: the minimum of the vertices.y
maxx: the maximum of the vertices.x
maxy: the maximum of the vertices.y

And I think it would probably be easy to also do all of these points in theta rho, although I am not sure what the use case would be.

In python, I would use curly braces {} because it would be easy to implement. IDK if there is a similarly easy string replacement in javascript.

jeffeb3:
Fixes #154. I pasted an example of what it looks like there.

I managed to get theta/rho to work without much trouble.

Some thoughts on this:
1. This is a completely undocumented feature. There isn't a good way for anyone to stumble into this. I am seriously considering making documenation for sandify. It is getting pretty complicated, and a little bit of documentation could go a long way. I would love to give some examples of good pre/post for people with sand tables, or people who want to use their CNC machine to plot with these patterns.
2. I am not sold on the naming scheme of these. It is just a case insensitive string. So I could have also done `{start x}`, which would also match with `{Start X}`. Or `start_x`, or `start-x`.
3. There is a lot of repetition in this code, but I couldn't think of a better way to do it. Maybe there is a more ES6/Javascript way to do this.

This is sort of a "vendetta" feature for me. Because I want to be able to plot a bunch of these neat patterns we've made, and I'd like to do something like this:

```
G1 Z10 F400 ; pen up
; Trace the boundary, so you can check for paper alignment
G1 X{minx} Y{miny} F2000
G1 X{maxx} Y{miny}
G1 X{maxx} Y{maxy}
G1 X{minx} Y{maxy}
G1 X{minx} Y{miny}
; Go to the start
G1 X{startx} Y{starty}
G1 Z-0.1; pen down
G1 F2000 ; set speed
```

### DIFF
--- a/src/features/exporter/Exporter.js
+++ b/src/features/exporter/Exporter.js
@@ -5,16 +5,13 @@ export default class Exporter {
   }
 
   export() {
+    this.pre = this.props.pre
+    this.post = this.props.post
     let vertices = this.props.vertices
     if (this.props.reverse) {
       vertices = vertices.reverse()
     }
-
-    this.pre = this.props.pre
-    this.post = this.props.post
-    if (vertices.lenth !== 0) {
-      this.variableReplace(vertices)
-    }
+    this.computeOutputVertices(vertices)
 
     this.header()
     this.startComments()
@@ -34,7 +31,7 @@ export default class Exporter {
     }
 
     this.line()
-    this.exportCode(vertices)
+    this.exportCode(this.vertices)
     this.line()
 
     if (this.post !== '') {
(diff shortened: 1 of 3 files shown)


bobnik:
Looks good to me. When we update the wiki these placeholders can be documented and we could add a link on the export dialog.

PR #167, opened 2020-08-03
Feature/file import

bobnik:
Originally a PR on my repo, now a PR vs master (because I accidentally approved my PR early).


jeffeb3:
Not a fan of the Font Awesome new/import :) ? OK. I think it does look good this way. :+1: 

This looks good to me. Are you finished (with this PR)?

jeffeb3:
Here's a bug. If I make a second layer, with zero vertices, I get an error sprouting from machine/selectors.js:144. I can make a layer with no vertices by setting numLoops to 0.

```
Uncaught TypeError: vertex is undefined
    offset geometry.js:99
    makeGetPreviewVertices selectors.js:144
    makeGetPreviewVertices selectors.js:143
```

I maybe don't understand javascript enough. Shouldn't the map just not call the function at all if there is nothing in the list? Is the problem that vertices is undefined, and not an empty list? Or is it a list with a single `undefined` in it?

jeffeb3:
Ah, so the next layer's vertices were undefined, and then we called cat, which made a list with an undefined inside.

Good. The map isn't breaking my world view :).

jeffeb3:
I have to quit for tonight. I just noticed if I set up the Space Filler and then hit the restore defaults I get a "machine is null" error. This must be my fault, but usesMachine is set to true for the space filler, so I'm not sure.

bobnik:
> Not a fan of the Font Awesome new/import :) ? OK. I think it does look good this way. 👍
> This looks good to me. Are you finished (with this PR)?
I think the small buttons hide the features, and they are very important ones. That, plus the fact that we only have 2 actions to take there and real estate is not an issue, made me decide to change it. 

Not finished yet. Will aim to finish today.

bobnik:
> I have to quit for tonight. I just noticed if I set up the Space Filler and then hit the restore defaults I get a "machine is null" error. This must be my fault, but usesMachine is set to true for the space filler, so I'm not sure.

You had the setting backwards. I pushed a fix.

jeffeb3:
I'm good with this again. Can you check that last fix I pushed? The current layer track was being drawn on every layer. I wasn't sure if the makeGet pattern was for pushing the layerId down, or if it was because of some caching choice, but I converted the getPreviewTrackVertices to that pattern to get the layerId passed in.

bobnik:
> I'm good with this again. Can you check that last fix I pushed? The current layer track was being drawn on every layer. I wasn't sure if the makeGet pattern was for pushing the layerId down, or if it was because of some caching choice, but I converted the getPreviewTrackVertices to that pattern to get the layerId passed in.

Good catch. I'd say we're ready to merge. Will leave for you.

PR #169, opened 2020-08-05
set default shape on page reload to be the same as the last one added…

bobnik:
… (or polygon if none)

"currentShape" was broken with the multiple layers branch merge. Re-adding it...


jeffeb3:
The types and the names aren't always the same. So this was reading in "clover" from the local state, and then trying to find that, instead of "epicycloid".

PR #175, opened 2020-08-12
Feature/layer refinements

bobnik:
Adds new "fine tuning" settings (at very bottom of properties) that allow a user to make fairly sophisticated adjustments when transitioning between layers. The inspiration for this PR is an upcoming Sandify pattern competition. When traversing layers, the ball often has to travel in visually unattractive ways. Some settings to remedy that:

- connectionMethod: 'line' (draw line between layers - what we do now), 'along perimeter' (walk along perimeter when connecting shapes)
- backtrackPct: After drawing path, move backward along pattern
- drawPortionPct: Only draws a portion of the pattern.
- rotateStartingPct: Treat the pattern as a loop, and rotate the starting point of the pattern before drawing.

In addition, I added a polar machine export setting polarMaxRho which defaults to 1, but can be reduced. This is because the ball on my Sisyphus table can get stuck. I'm not 100% sure why it happens, but the table translates a rho=1 into a distance > 1 and the ball bumps against a plastic wedge that is used to hold a second (optional) ball. I will pursue troubleshooting with Sisyphus, but long story short, setting a rho max of .99 creates an identical pattern that is able to complete. I believe others with this table will have similar issues, and so the setting feels like a valid addition to Sandify.

With fine tuning, in this pattern I only have to draw across the layers twice in the entire pattern, and in both cases, it's just tiny lines.

image


src/common/util.js
```diff
@@ -21,3 +21,11 @@ export const reduce = (numerator, denominator) => {
   gcd = gcd(numerator, denominator)
   return [numerator/gcd, denominator/gcd]
 }
+
+// rotates an array count times
+// taken from https://stackoverflow.com/questions/1985260/rotate-the-elements-in-an-array-in-javascript#33451102
+export const arrayRotate = (arr, count) => {
```
jeffeb3, line 27: If you have the points 0,1,2,3. If you rotated it, you would end up with 2,3,0,1. But the line
segment from 3,0 wasn't in the original pattern, and the segment 1,2 is broken and will no longer be
drawn.

Even though the vectors are a list of points, the pattern is really a set of line segments. This is
messing up a little by breaking the line segment between the new end and the new start.

I was thinking of building a list of line segments, and then rotating them, and then compressing
them, but I think it's probably something simpler. In fact, I think we should be able to just insert
the new final point at the very beginning.

Dealing with 0,3 is harder. That may just need to be something that the user has to deal with.

src/features/app/store.js
```diff
@@ -31,9 +31,15 @@ const store = configureStore({
 // settings across page loads; don't forget to toggle false when done testing!
 const persistState = false
 
+// if you want to save a multiple temporary states, use these keys. The first time
+// you save a new state, change persistSaveKey. Make a change, then change
+// persistInitKey to the same value. It's like doing a "save as"
+const persistInitKey = 'state'
+const persistSaveKey = 'state'
```
jeffeb3, line 38: Can you change these things in the developer console?

src/features/exporter/Downloader.js
```diff
@@ -153,20 +154,28 @@ class Downloader extends Component {
               index={1}
               model={this.props} />
 
+            {this.props.fileType === 'Theta Rho (.thr)' && <InputOption
```
jeffeb3, line 157: Should we just always multiply by 0.99? Is that going to mess people up?

src/features/layers/Layer.js
```diff
@@ -130,7 +130,7 @@ class Layer extends Component {
 
         <div className="pt-1">
           { optionsListRender }
-          { this.props.layer.canTransform && <Transforms id={this.props.layer.id} />}
```
jeffeb3, line 133: What am I missing here? Why remove this check?

src/common/util.js
jeffeb3: n/m. I see you handle this with the flag to close the loop. What shapes get worse if you always insert that point? I was trying this with the wiper (lines) and it was pretty easy to get it to miss a big line.

src/features/machine/rectMachine.spec.js
```diff
@@ -68,4 +69,13 @@ describe('rect machine', () => {
       expect(machine.tracePerimeter({x: 250, y: -50}, {x: -250, y: -50})).toEqual([{x: 250, y: -250}, {x: -250, y: -250}])
     })
   })
+
+  describe('nearestPerimeterVertex', () => {
+    it('finds the nearest vertex', () => {
```
jeffeb3, line 74: :100: Always like to see unit tests like this. These will keep us honest :)

src/features/layers/Layer.js
jeffeb3: Ah, you moved the check into the layers. :heavy_check_mark:

jeffeb3 (approved):
Looks like some solid improvements.

src/common/util.js
bobnik: You are right. Rotation does not work for all shapes. It's mainly to move the ending point in certain circumstances to a place where it can connect to the next layer. It's up to you, but I'm ok with this not always working given it's a "fine tuning (advanced)" setting.

src/features/app/store.js
bobnik: Not at the moment. This will largely go away once we support Sandify export format.

src/features/exporter/Downloader.js
bobnik: Almost all patterns work fine. Some don't (on my table). I wouldn't change the default from 1 for sure. If you think this feature feels awkward, we can hide it so that I can use it when needed for personal use. Sisyphus forums show this happens to others, but their reply is usually along the lines of "don't move the ball so fast, the tracking can get off". I have not found this to be true in all cases, but I don't know the cause.

PR #184, opened 2020-09-20
Height and width

bobnik:
Allows independent sizing of width and height.


src/features/preview/PreviewWindow.js
```diff
@@ -131,7 +131,8 @@ class PreviewWindow extends Component {
               e.evt.preventDefault()
               if (Math.abs(e.evt.deltaY) > 0) {
                 this.props.onLayerChange({
-                  startingSize: scaleByWheel(this.props.currentLayer.startingSize, e.evt.deltaY),
+                  startingWidth: scaleByWheel(this.props.currentLayer.startingWidth, e.evt.deltaY),
```
jeffeb3, line 134: This is adding a number to each of them. So if I start with 5,10, it scales to 10,15, not 10,20.

I did a little bit of code to scale the height based on the difference in width, but it ends up with fractions. If I clip it with round, it ends up getting off over lots of scrolls. 

Not sure what the best solution is, but I'm ok with how it works now, because it is at least undoable. But it approaches square as the scale gets larger.

src/models/space_filler/SpaceFiller.js
```diff
@@ -50,6 +50,7 @@ export default class SpaceFiller extends Shape {
         fillerSubtype: 'Hilbert',
         repeatEnabled: false,
         canChangeSize: false,
+        autosize: false,
```
jeffeb3, line 53: We probably need this on the new noise pattern too.

PR #185, opened 2020-09-20
noise wave shape

bobnik:
This pattern is based on perlin (and simplex) noise. It produces endless, attractive (to me) patterns.

This shape has a few "rough edges". When you drag the shape in the editor, you see some artifact perimeter lines that go away as soon as you stop dragging. Additionally, some random seeds produce curves with the same artifact lines that mar the pattern. This is due to the way I bound the curves, and I didn't see any easy way around it. Regardless, given that most patterns look great, I feel like this is acceptable. Feedback or suggestions to fix obviously welcome.

image


jeffeb3:
Is it possible to publish the noise.js in another repo? I'm a little cautious about licenses. I see this is published with an MIT license, so it may not be a huge deal. But it might also be easy to make a bobnik/perlinnoise and then be able to npm install it like our other deps.

jeffeb3:
screenshot-2020-10-02-1601651439

There's some line noise here. Since master has the advanced features installed, this might be because of some of that math (but I left those at default). I am wondering if one of the lines ended up in the middle and it's taking off because of it.

jeffeb3:
FYI, those are with the default settings, but with the machine size set to 500. If I crank up the iterations to 500, they go away. I am still looking into the math to try to figure out how the particles are connected, but I think they aren't travelling all the way off, so they are just jumping to the next line.

bobnik:
Yes, that's absolutely what happens. I didn't find an efficient way to
eliminate all lines across all seed and machine sizes.

jeffeb3:
What about eliminating particles that don't end outside the perimeter? It seems like they are all the particles right in the middle, where there are other lines nearby.

jeffeb3:
Regarding the drag and drop funkiness. What about doubling or 10xing the machine limits math, so the outer ring is just really far away? Haha, That might be a problem for the other solution I just posted.

jeffeb3:
Or maybe the max bound on the number of iterations should be based on the line reaching the perimeter? I am sure there are some very small numerators at some point we would still have to drop.

jeffeb3:
That actually seems to be working. I will push what I have soon.

jeffeb3:
The previous curve similarity thing is only looking at the very last curve. Since they come in random order, two similar curves won't be that nearby, right?

Is there a good way to "sort" them in a way that would make similar particles with similar curves be nearby? That would be a neat addition to the pattern anyway.

I am thinking that maybe until we have a way to do that, we should just cut the nearby calculation all together.

jeffeb3:
I played with it without the curve matcher and it seems ok to me. I didn't do any profiling, so I'm not sure it really hurts us to leave it in.

Your call. It can easily be reverted.

From my checks, this seems ready to merge.

bobnik:
> I played with it without the curve matcher and it seems ok to me. I didn't do any profiling, so I'm not sure it really hurts us to leave it in.

I am just now getting back to this. Sorry for the delay. I have looked over your changes. Some comments:

- "Pattern number" vs "Random seed". I like random seed, especially given these algorithms are all about pseudo-randomness. How about "Random seed (changes pattern)"?
- Removal of iterations setting. I used this to good effect during the contest to generate minimal patterns. That said, I can see that auto-generation consistently builds nice patterns, so this seems okay.
- Removal of curve similarity. I think this should be added back in. WIthout it, you end up with patterns that draw almost the same lines over and over again, which is boring for the observer and creates deep rivets in the sand. As example for a random seed, at 88% distance  = 80356, at 83% with a very similar macro appearance, distance = 58910. It's an advanced setting, for sure, but it's a useful one. If you disagree, a fallback is to hardwire a similarity check in the code (say, 85%) and hide the option from the user.

jeffeb3:
This has a hard coded curve matcher of 88. If we have to put it in as an option, I'd like to have it be a better scale. It seems like 80-90 are basically all the useful values. So maybe we should scale it so that ends up being a more intuitive range.

PR #189, opened 2021-01-01
tiny package change to fix install dependencies

bobnik:
I submitted a more ambitious PR earlier to do updates, but react-select does not play nice with it and I couldn't easily fork and resolve. We may have to replace react-select soon to progress with React 17 and beyond.


bobnik:
Merged these changes into #190, so closing this PR...

Issue #182, opened 2020-09-16
Layer manipulators


jeffeb3:
There is a lot of neat things we can do with layer manipulators. Like a filter that acts on the previous layer. The transforms is kind of like that, and we could probably force the functionality into more transforms. But it might also make sense to have some layers who's only job is to manipulate the layer above.

I know Bob has started this with some limiting code, to be able to limit the previous layer to a region using the machine limits code.

I'd also like to be able to add waves, or skew/stretch patterns. Or just a general f(x,i), f(y,i) transforms.

Needs some thought. Should these "manipulators" be new layers, or should they be added to the transform area? If we make them layers, they would be easier to compound.

Should the existing transforms be converted into manipulators?

Should the transform area just be completely converted into a different kind of interface?

bobnik:
I've been working on a mask layer which works as a "manipulator". At the moment, it's a hard-wired special case that doesn't draw itself, but instead acts on the previous layer (as you describe). It's also a layer itself, which makes it easy to drag and drop it wherever you want it. This may not ultimately be how we treat this type of feature (e.g., it could attach to a layer instead of being one), but I'd like us to play around with it a bit before deciding. I can definitely see layer manipulators being made more generic.

I'm not convinced that all existing transforms should be fit into this construct, mostly because it would effectively hide a number of cool functions from casual Sandify users. At the moment, you pick a shape and start playing. As a manipulator, you'd have to be aware of its existence and then add it to your pattern.  The upside of this separation, though, would be that you could easily take your transforms and copy or move them. Maybe we could eventually move advanced transforms like track into a separate manipulator layer.

One other consideration is serialization. Ideally, we'll want to make any "big" data model changes before we support import/export of a native Sandify format, per our roadmap.

jeffeb3:
I agree. I think the original intent of this issue has been solved by adding the effects. I would like to see what else we can do in an effect. I think we have a lot of power now to do some fun skew/stretch/limiting. It would be fun to try to do some wavy distortion on something like the circle loops.

But this issue is "finished". For any new ideas, PRs are welcome.

PR #190, opened 2021-01-01
Layer masks

bobnik:
This PR allows you to add one or more circle or rectangle mask layers after a shape that clip the boundaries of that shape after other transforms are run. It takes our machine boundary logic and lets you apply it within your pattern, with great flexibility. I wrote most of this code to create interesting submissions for the recent Sisyphus contest.

After digesting #182, I decided to make the Mask shape an "effect" (renamed from manipulator to mimic Photoshop). Effects are treated a bit differently that other shapes. They have fewer options and can't be changed into other shapes after creation. They are also displayed under their parent shapes in the play list. Down the road, we can add other effects if we don't want to include them in every shape by default.

image

To test:
- Add a new shape.
- Add a mask below it. 
- Play around moving the shape around, and also moving and re-sizing the mask.
- Try out the various mask options.


jeffeb3:
I think it would be useful to draw the perimeter of the mask when the mask layer is selected. The square has the konva handle, but the circle is harder to "feel".

jeffeb3:
screenshot-2021-01-03-1609688684

src/features/preview/MachinePreview.js
```diff
@@ -41,7 +41,7 @@ class MachinePreview extends Component {
               <div className="p-3">
                   <Slider
                     value={this.props.sliderValue}
-                    step={0.1}
+                    step={0.05}
```
jeffeb3, line 44: :100:

src/features/preview/PreviewLayer.js
```diff
@@ -275,8 +275,9 @@ const PreviewLayer = (ownProps) => {
           ref={trRef}
           centeredScaling={true}
           resizeEnabled={!props.layer.trackEnabled}
+          rotateEnabled={props.layer.maskMachine !== 'circle' }
```
jeffeb3, line 278: Should we be pushing these into the layer/shapes? Some default and then the circle mask would be doing something unique, instead of putting circle mask specific code here?

src/features/transforms/Transforms.js
```diff
@@ -144,7 +144,7 @@ class Transforms extends Component {
         </Accordion>
       }
 
-      <Card className="mt-3">
+      {!this.props.layer.effect && <Card className="mt-3">
```
jeffeb3, line 147: This isn't how we handle disabling other features (like the transforms).

src/models/Mask.js
```diff
@@ -0,0 +1,94 @@
...
+    ]
+  }
+
+  applyEffect(effect, layer, vertices) {
+    vertices = vertices.map(vertex => {
+      return rotate(offset(vertex, -effect.offsetX, -effect.offsetY), effect.rotation)
+    })
+
+    if (!layer.dragging && !effect.dragging) {
+      const machineClass = effect.maskMachine === 'circle' ? PolarMachine : RectMachine
```
jeffeb3, line 77: I really like this reuse. I was afraid of how this would work.

src/common/geometry.js
```diff
@@ -126,3 +126,15 @@ export const getSliderBounds = (vertices, sliderValue) => {
 
   return { start: start, end: end }
 }
+
+// returns an array of points drawing a circle of a given radius
+export const circle = (radius=1, start=0) => {
+  let points = []
+
+  for (let i=start; i<=128+start; i++) {
+    let angle = Math.PI * 2.0 / 128.0 * i
```
jeffeb3, line 135: You can switch to counterclockwise by multiplying i by -1. I don't think this matters in the mask, but...

jeffeb3:
This is really excellent and it works very smoothly. I have some comments, but nothing that will hold this up.

src/features/layers/Playlist.js
```diff
@@ -79,15 +79,17 @@ const customStyles = {
...
   const dragClass = canRemove ? 'cursor-move' : ''
+  const nameClass = effect ? 'no-select font-italic': 'no-select'
+
   return <ListGroup.Item className={[activeClass, dragClass, 'd-flex align-items-center'].join(' ')} key={id} id={id} onClick={onLayerSelected}>
     <Button className="layer-button" variant="link" data-id={id} onClick={onToggleLayerVisible}>
       {visible && <FaEye />}
       {!visible && <FaEyeSlash />}
     </Button>
-    <div className="no-select">{name}</div>
+    <div className={nameClass}>{effect && '\u2014'}{name}</div>
```
jeffeb3, line 92: How about U+2937 instead? ⤷

bobnik:
> This is really excellent and it works very smoothly. I have some comments, but nothing that will hold this up.

I will do some cleanup and we can merge. Can we consider a new release after this? We have a number of new features.

jeffeb3:
Yes, definitely. It is about time. Sometime this week.

src/features/transforms/Transforms.js
bobnik: Not sure I understand this...

bobnik (commented):
I've committed some edits to address most of the above. I added code to outline a circular mask when dragging it, but not when stationary. Hopefully that gets us close enough - if not, it will require more refactoring to not draw connectors to the mask and to somehow differentiate between a bordered mask (setting) or just an outlined mask.

jeffeb3:
This works great.

PR #191, opened 2021-01-08
allow insertion of new layer after current one

bobnik:
Allows insertion of a new layer after the current one, which is something that becomes quite useful when you start having a large number of layers. Prior to this PR, you have to carefully drag each new layer to the right place.

### DIFF
--- a/src/features/layers/Playlist.js
+++ b/src/features/layers/Playlist.js
@@ -1,6 +1,6 @@
 import React, { Component } from 'react'
 import { connect } from 'react-redux'
-import { Accordion, Button, Card, ListGroup, Modal, Row, Col, Form } from 'react-bootstrap'
+import { Accordion, Button, Card, ListGroup, Modal, Row, Col, Form, ToggleButton, ToggleButtonGroup } from 'react-bootstrap'
 import Select from 'react-select'
 import { SortableContainer, SortableElement } from 'react-sortable-hoc'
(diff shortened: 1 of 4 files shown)


jeffeb3:
The selection buttons are a bit clunky. I don't mind changing the behavior to always insert after the current layer, or always at the end. I think it makes most sense to do it after the current layer. I don't think it is that much work to move it if it is in the wrong place, and I don't think it is much work to just delete it and make it new again if it isn't in the wrong place.

bobnik:
> The selection buttons are a bit clunky. I don't mind changing the behavior to always insert after the current layer, or always at the end. I think it makes most sense to do it after the current layer. I don't think it is that much work to move it if it is in the wrong place, and I don't think it is much work to just delete it and make it new again if it isn't in the wrong place.

I had patterns with 30+ layers and trust me when I say it's not easy to move them from the end to their "correct" place. That said, if you find the option unnecessary I'd prefer "after current" to "at end" and could make that change instead.

jeffeb3:
Fair enough. I think if you are in the situation where you have a bunch of layers and you're trying to move them up, then deleting the layer and selecting the right layer to insert it is not too hard. And if you are in the situation where you are really making a lot of choices in the new or import dialogs, then you would prefer to move it. If you were in the awful situation where you just inserted it in the wrong place and you have 30+ layers and the import/new dialog is very complicated, then... That's a bummer.

bobnik:
Obsolete, after #193...

PR #193, opened 2021-01-11
when adding a new layer, insert after current one (vs at end)

bobnik:
Pared down version of #191. When adding new layer, add after current one, not at end.


jeffeb3:
This still doesn't work the way I was expecting w.r.t. copies. It will always select the layer that you are trying to copy.

PR #195, opened 2021-01-16
Feature/selector optimization

bobnik:
This PR is a selector cleanup effort. In some cases, our selectors are overly broad in their selection, resulting in unnecessary recalculation. I fixed some of the more obvious cases. There are a number of places where we still do expensive calculations repeatedly, and I could not figure out how to resolve. One such issue is the mix of connected Redux components and a useSelector hook (due to Konva interaction). As I do not anticipate refactoring to remove this hybrid approach any time soon, I expect we can live with some inefficiency.

I was unable to observe a visible increase in performance when using the UI. That said, I was concerned about the unlikely possibility of _slowing things down_. I tested with a large import and it felt about the same. Obviously, please do your own spot testing.

### DIFF
--- a/src/features/exporter/CommentExporter.js
+++ b/src/features/exporter/CommentExporter.js
@@ -1,5 +1,4 @@
 import { getShape } from '../../models/shapes'
-import { getLayerInfo } from '../layers/selectors'
 import Machine from '../../models/Machine'
 import Transform from '../../models/Transform'
 import Exporter from './Exporter'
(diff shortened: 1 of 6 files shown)


jeffeb3:
The line connecting layers seems to be missing in this branch.

jeffeb3:
I can't find anything else different from the testing, and the code looks good to me.

bobnik:
> The line connecting layers seems to be missing in this branch.

Ah, you're right. An oversight, and worse, I don't know how to fix. I'll revisit later and let you know.

bobnik:
Will resubmit this as a separate PR.

PR #196, opened 2021-01-17
Fisheye effect

bobnik:
A new fisheye effect deriving from the Effect class (like Mask)!

image


src/models/Fisheye.js
```diff
@@ -0,0 +1,58 @@
...
+  getVertices(state) {
+    const width = state.shape.startingWidth
+    return circle(width/2, 0)
+  }
+
+  applyEffect(effect, layer, vertices) {
+    const fisheye = d3Fisheye.radial()
+      .radius(effect.startingWidth/2)
+      .distortion(effect.fisheyeDistortion/2)
+    fisheye.focus([effect.offsetX, effect.offsetY]);
```
jeffeb3, line 47: Cool!

jeffeb3:
This is really fun. And I will merge this as is. I think this is upping the importance of having a function to break up the longer line segments into smaller pieces. This effect doesn't work on the polygons, for example. We need a way to just pick a number, like 1mm and have each line segment turned into enough line segments that there won't be any lines longer than 1mm.

jeffeb3:
screenshot-2021-01-23-1611426362

Reuleaux has bumps!

bobnik:
> This is really fun. And I will merge this as is. I think this is upping the importance of having a function to break up the longer line segments into smaller pieces. This effect doesn't work on the polygons, for example. We need a way to just pick a number, like 1mm and have each line segment turned into enough line segments that there won't be any lines longer than 1mm.

@jeffeb3 I'd be hesitant to do this automatically only because would add a bunch of extra vertices and could affect performance. Perhaps we only break it up when certain effects are used (e.g., fisheye)

jeffeb3:
Yeah. I agree. I am also thinking we may want to have some form of super sampling lines when they are nearly straight for certain operations. I'd like to be able to do as much of this automatically as possible, but I know there are limits on my time to be able to mess with this.

jeffeb3:
There are a few other places. I can specifically think of the preview slider, and some of the advanced tuning options. You can't do some of the advanced tuning stuff because a square only has 4 places.

Some of those might be better handled by more clearly handling fractions of loops.

PR #197, opened 2021-01-23
Feature/scara

jeffeb3:
@bobnik 

There are a few people in the forums that are trying to create polar sand table machines using scara type mechanisms. I think the ultimately correct choice would be to have the firmware be smart enough to handle this kind of geometry, but that has proven difficult enough that no one has been successful with that.

So I made a workaround:
https://forum.v1engineering.com/t/sandtrails-a-polar-sand-table/16844/61?u=jeffeb3

But even though I tried to make the converter simple enough to use, I know people would prefer to have this done in sandify.

Even though it isn't very hard, I am worried it will be a distraction. WDYT? I still need to make a long page of documentation for it in the wiki. It really is sort of like me making a brand new standard file type.


jeffeb3:
I guess I need to write up this wiki.

jeffeb3:
I want to add a parameter for the units/circle.

I want to write up at least some wiki information before we release this (it can be moved to master before that though).

PR #199, opened 2021-01-28
Feature/function effect

jeffeb3:
This is fun. I was playing with the Custom Warp functions and I created some neat patterns, so I parameterized them and they are even more fun to play with, and harder to mess up.

This star has a 'circle' warp type:
star_warp

This wiper has a 'angle' warp type:
wiper_warp

This square has the 'quad' warp type:
polygon_quad_warp


jeffeb3:
'grid':

screenshot-2021-01-28-1611870311

bobnik:
@jeffeb3 I don't think this needs to hold up the merge, but I do think x-shear and y-shear should be separate. The visual effect is different for each. The combined shear seems odd to me.

jeffeb3:
I may not have the signs right. But rotating it 90 degrees should be the Y-Shear.

bobnik:
> I may not have the signs right. But rotating it 90 degrees should be the Y-Shear.

Duh. Got it :)

PR #200, opened 2021-02-06
scara PR suggestions

bobnik:
I felt that the Scara GCode format could be extracted more cleanly into a new exporter, and then got a bit carried away with refactoring the exporter constants. Hence the PR... if you don't like the refactor, take what you like out of it manually.


jeffeb3:
Thank you. This is great stuff.

PR #201, opened 2021-02-06
function-effect PR suggestions

bobnik:
This opinionated PR to your PR suggests:

- moving custom into warp (with appropriate options being shown)
- changing name to '2D Transformation'
- adding new x-shear and y-shear options
- making the scale option starting width instead so you can use the mouse to scale up and down. Otherwise, the starting width does nothing.

One other thing I noticed that I did not try to figure out... trying adding a warp effect to a shape, then rotate the shape (not the warp). The rotation does not seem to work properly. Ideally, the rotation of the shape (including the effect) would be entirely independent of the rotation of the effect, if this make sense...


jeffeb3:
> making the scale option starting width instead so you can use the mouse to scale up and down. Otherwise, the starting width does nothing

jeffeb3:
> changing name to '2D Transformation'

jeffeb3:
> The rotation does not seem to work properly

Maybe the warp should have the option for rotating on it's own? I haven't tried to use this in a pattern. I can't decide which is more useful. I can imagine wanting to tweak the shape and the rotation together, but it also seems intuitive to make them rotate independently, so you can fine tune each of them.

jeffeb3:
ok. I don't like the bounding circle, and I like the scale being related to the width. I didn't see an easy way for the initial width to not be shown, so I just removed the scale and we will have to remember that Initial width is controlling the strength of the warp.

jeffeb3:
I also combined the two shears, and let it rotate.

bobnik:
> > changing name to '2D Transformation'
> 
> I like warp better :). I thought that was your name for it?

Yeah, it was :) I suppose I thought "2D transformation" was more accurate, but I do like "warp". We've always leaned toward user-friendly/fun vs. mathematical naming, so "warp" is consistent with that.

PR #202, opened 2021-02-07
Feature/inverted mask

bobnik:
image

This PR introduces the concept of an inverted mask (setting on the Mask shape). It works for both polar and rect masks, and can be nested with other masks to create some interesting clipping effects. I've done a reasonable amount of testing of the mask with different shapes, but please put it through its paces.

I've also added a Freeform shape, which I used to test the new machine logic. It's not really intended for end-user usage, so I've commented it out, but I can see an enhanced version of something like it eventually becoming an end-user shape.

Lastly, I made small corrections to the original boundary logic - hopefully I didn't break anything.

### DIFF
--- a/src/features/layers/layersSlice.spec.js
+++ b/src/features/layers/layersSlice.spec.js
@@ -41,7 +41,6 @@ describe('layers reducer', () => {
     backtrackPct: 0,
     drawPortionPct: 100,
     effect: false,
-    rotateCompleteLoop: false,
     rotateStartingPct: 0,
     growEnabled: true,
     growValue: 100,
--- a/src/features/machine/computer.js
+++ b/src/features/machine/computer.js
@@ -156,18 +156,6 @@ export const transformShapes = (vertices, layer, effects) => {
     vertices.pop()
   }
 
-  if (layer.rotateStartingPct === undefined || layer.rotateStartingPct !== 0) {
-    const start = Math.round(vertices.length * layer.rotateStartingPct / 100.0)
-    vertices = arrayRotate(vertices, start)
-
-    // some patterns create a loop when drawn and transformMethod is 'intact'; in
-    // this case, rotateStartingPct will not draw the final vertex to complete the
-    // loop. Use rotateCompleteLoop to add it back in for specific shapes.
-    if (layer.rotateCompleteLoop) {
-      vertices.push(new Victor(vertices[0].x, vertices[0].y))
-    }
-  }
-
   if (layer.trackEnabled && numTrackLoops > 1) {
     for (var i=0; i<numLoops; i++) {
       for (var t=0; t<numTrackLoops; t++) {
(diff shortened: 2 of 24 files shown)


bobnik:
Jeff, there are still some boundary condition issues. You can see it is you try to use a noise wave (both with fixes I put into the polar machine and with the new inverted machines. I'll try to fix them, and obviously please check out the PR in general, but this is not yet ready to merge.

bobnik:
@jeffeb3 Perlin+mask issue is fixed, and this PR is again ready for review. Thanks.

src/features/layers/layersSlice.spec.js
```diff
@@ -41,7 +41,6 @@ describe('layers reducer', () => {
     backtrackPct: 0,
     drawPortionPct: 100,
     effect: false,
-    rotateCompleteLoop: false,
```
jeffeb3, line 44: We needed this to fix some bug, right? Is this just a hack that we don't need anymore? The default for all shapes is to have it set to false, but then several shapes turn it on.

src/features/machine/computer.js
```diff
@@ -176,16 +164,27 @@ export const transformShapes = (vertices, layer, effects) => {
...
-      for (var j=0; j<completion; j++) {
+      for (var j=0; j<vertices.length; j++) {
         let amount = layer.transformMethod === 'smear' ? i + j/vertices.length : i
         outputVertices.push(transformShape(layer, vertices[j], amount, amount))
       }
     }
   }
 
+  if (layer.rotateStartingPct === undefined || layer.rotateStartingPct !== 0) {
+    console.log('here!')
```
jeffeb3, line 175: Probably don't want these around anymore (hopefully).

src/models/Freeform.js
```diff
@@ -0,0 +1,43 @@
...
+  constructor() {
+    super('Freeform')
+  }
+
+  getInitialState() {
+    return {
+      ...super.getInitialState(),
+      ...{
+        type: 'freeform',
+        freeformPoints: '-1,-1;-1,1;1,1',
```
jeffeb3, line 24: What if choosing the freeform shape opened a modal (like the import) with a little window you could click into and make a new shape? It wouldn't be editable (unless we could add a button to reopen the modal). And you wouldn't be able to edit it, while looking at it's output transformed, warped, looped. But this seems like an easier way to add this feature.

Regardless, this should probably be in a separate PR.

jeffeb3:
screenshot-2021-03-04-1614918081

This is honestly a deceptively hard problem. This kind of looks like it isn't picking the right corner/side on the perimeter to return from, and it is picking a spot across the shape, and then returning through the middle.

I would be happy to merge/publish this (it is a lot better than nothing), but we will want to fix that eventually. I don't have time tonight (and I really need to focus on this to really get into it).

src/models/Freeform.js
bobnik: I like that. As it stands now, it's useful for debugging boundary conditions (it was invaluable when coding the inverted masks), but it's commented out of general use.

src/features/layers/layersSlice.spec.js
bobnik: I don't think we need it any more. I tested each affected shape, but there's a chance I missed something obviously.

bobnik:
> This is honestly a deceptively hard problem. This kind of looks like it isn't picking the right corner/side on the perimeter to return from, and it is picking a spot across the shape, and then returning through the middle.
 
I _think_ this is fixed. I made a change to cleanVertices, among others, which I'd like you to look at. I'm not clear why we had the logic we did in there.

bobnik:
@jeffeb3 the latest commit is an attempt to fix issue #204. It should also help with boundary conditions in general. LMK if you see any issues.

PR #203, opened 2021-02-07
use arrow keys to change offsets in machine preview

bobnik:
If the machine preview has focus, you can use arrow keys to move the current layer around. Shift makes it more fine-grained. I tried allowing arrow keys to work even outside of the machine preview, but it was interfering with other controls making use of them (like scrollbars or option fields), so I reverted that behavior.

I also removed the logic to allow deselection of the current layer by clicking on it. I seem to recall us talking about doing that, anyway, and it makes the movement using keys more consistent (you can't accidentally deselect a layer when clicking on the preview).

### DIFF
--- a/src/features/preview/MachinePreview.js
+++ b/src/features/preview/MachinePreview.js
@@ -4,12 +4,18 @@ import Slider from 'rc-slider'
 import 'rc-slider/assets/index.css'
 import PreviewWindow from './PreviewWindow'
 import Downloader from '../exporter/Downloader'
+import { getCurrentLayer } from '../layers/selectors'
+import { updateLayer } from '../layers/layersSlice'
 import { updatePreview } from './previewSlice'
 import { getVerticesStats } from '../machine/selectors'
 import './MachinePreview.scss'
 
 const mapStateToProps = (state, ownProps) => {
+  const current = getCurrentLayer(state)
+
   return {
+    currentLayer: current,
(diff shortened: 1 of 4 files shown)


src/features/preview/MachinePreview.js
```diff
@@ -20,13 +26,44 @@ const mapDispatchToProps = (dispatch, ownProps) => {
...
       dispatch(updatePreview({sliderValue: value}))
     },
+    onLayerChange: (attrs) => {
+      dispatch(updateLayer(attrs))
+    },
+    onKeyDown: (event, currentLayer) => {
+      let attrs = { id: currentLayer.id }
+
+      if (['ArrowDown', 'ArrowUp', 'ArrowLeft', 'ArrowRight'].includes(event.key)) {
+        const delta = !!event.shiftKey ? 1 : 5
```
jeffeb3, line 36: What is the !!? Is that not not?

The shift key seems to be highlighting text. I don't really mind, but I noticed in when I tried using shift+arrows.

jeffeb3:
LGTM. Merge when ready.

src/features/preview/MachinePreview.js
bobnik: !! converts the value to a Boolean.

jeffeb3: :+1:

PR #207, opened 2021-05-09
Feature/noise effect

bobnik:
Re-uses perlin and simplex noise functions in a layer effect. Works best over a simple repetitive pattern like wiper or polygon.

image


jeffeb3:
This is really neat. I really like the generated patterns.

1. It looks like there is a warning.
```
src/common/eulerianTrail.js
  Line 3:1:  Unexpected default export of anonymous function  import/no-anonymous-default-export
```
2. The subsample option is on the effect, but I think that it isn't doing anything
3. Are the options for "Application" useful? I could only find contour 1 to be useful. But maybe I am using them wrong? I haven't looked at the code yet, so I'm not sure what they mean.
4. If I choose different applications, it seems like the "Magnification" should be a different scale. If this means something physically, then ok. But is it possible to just apply a scale based on the application chosen to keep them all in a somewhat sane range? I can't seem to get anything pretty from the Linear one. The Contour 2 seems to look better with close to 10x the magnification.

Generally though, this is really neat. I haven't looked at the code yet. I will try to soon. Just been super busy.

jeffeb3:
This maybe should be a different issue. But it would be neat to have some scaling of the pattern, so the edges don't get pulled into the field of view. This is something that bugged me with the other effects too. Maybe at the core, it should just be changed in the wiper?

bobnik:
> This maybe should be a different issue. But it would be neat to have some scaling of the pattern, so the edges don't get pulled into the field of view. This is something that bugged me with the other effects too. Maybe at the core, it should just be changed in the wiper?

In the patterns I am generating, this doesn't happen every time and it's pretty minor when it does.

bobnik:
> This is really neat. I really like the generated patterns.
> 
> 1. It looks like there is a warning.
> 
> ```
> src/common/eulerianTrail.js
>   Line 3:1:  Unexpected default export of anonymous function  import/no-anonymous-default-export
> ```
> 
> 1. The subsample option is on the effect, but I think that it isn't doing anything
> 2. Are the options for "Application" useful? I could only find contour 1 to be useful. But maybe I am using them wrong? I haven't looked at the code yet, so I'm not sure what they mean.
> 3. If I choose different applications, it seems like the "Magnification" should be a different scale. If this means something physically, then ok. But is it possible to just apply a scale based on the application chosen to keep them all in a somewhat sane range? I can't seem to get anything pretty from the Linear one. The Contour 2 seems to look better with close to 10x the magnification.
> 
> Generally though, this is really neat. I haven't looked at the code yet. I will try to soon. Just been super busy.

- The warning is fixed in master.
- I removed 'Contour2', but left Linear, as I do think it has a place, even if it's not pretty. You can generate some interesting scratchy patterns.
- I tried to adjust magnification so that it's both pleasing at default value and also consistent between noise types. I also changed the wiper default size so applying noise to it is initially pleasing. I think the tighter pattern with the smaller default works in general, but feel free to adjust this if you disagree.

jeffeb3:
Today is the day. Looks good to me.

PR #209, opened 2021-06-12
Feature/selector optimization 2

bobnik:
This is a refactor of my earlier optimization PR. The primary goal is to optimize selector invalidation. Please test this extensively (specifically, a variety of layer manipulations, both within the layer and between (moving layers, show/hide layer, etc.)

A few other features:
- set debug = true in common/util.js to log selector invocations
- set markCoordinates = true in preview/PreviewLayer to see coordinates for end points


src/common/util.js
```diff
@@ -29,3 +29,9 @@ export const arrayRotate = (arr, count) => {
   arr.push.apply(arr, arr.splice(0, count))
   return arr
 }
+
+// set to true to turn on console logging
+const debug = false
+export const log = (message) => {
+  if (debug) { console.log(message) }
```
jeffeb3, line 36: If this is always false, does this just get "optimized out"? Nifty feature, if it did.

jeffeb3: I see that some of the other information, specifically the file/line are set to this log location, instead of the source file.

Does it make more sense to do something like `console.debug()`? I am not familiar with the nuances here. I wouldn't want it to be slower for the end user, but I also don't mind if someone reads our debug logs at sandify.org.

jeffeb3:
This seems to work great. I don't really understand it. I really need a refresher on redux, I guess.

jeffeb3:
screenshot-2021-09-01-1630517263

I was playing with a mask, on a star, with another layer there too. I was grabbing the resize corner of the mask (I think) and I got this error. I could easily check to see if props.layer.id is valid, but I don't understand it enough to know if this function should have even been called. I haven't been able to reproduce it. Do you have any thoughts on it?

src/common/util.js
bobnik: I don't know how much slower spitting out log messages would make pattern rendering, but on principle I don't see a reason to spam the logs unnecessary, especially if we're just using for debugging purposes when something isn't working. Reading up on console.debug, it seems to be the same as console.log, just hidden behind a filter. As far as line number goes, the log message itself tells you where you are. Anyway, my inclination is to leave it as-is. We can always use console.log (or debug) in places where things go wrong, like you have done in boundary condition logic.

bobnik:
> I was playing with a mask, on a star, with another layer there too. I was grabbing the resize corner of the mask (I think) and I got this error. I could easily check to see if props.layer.id is valid, but I don't understand it enough to know if this function should have even been called. I haven't been able to reproduce it. Do you have any thoughts on it?

I can't seem to reproduce this. I'm hesitant to introduce an issue that crashes the app into production, but I'm not sure how to fix if I can't reproduce. It's up to you. The optimizations will certainly help general use.

PR #210, opened 2021-06-20
Circle packer

bobnik:
image

This PR implements a circle packer shape. The algorithm to draw the circles was easy to implement. The algorithm to walk the circles was more challenging. 

Even now and then I see a border line error, but I think it's in the main machine logic, not the shape. That said, you can always adjust the parameters of the shape to make it go away, so I don't think it's a big issue.


src/common/geometry.js
```diff
@@ -128,17 +128,36 @@ export const getSliderBounds = (vertices, sliderValue) => {
...
 // returns an array of points drawing a circle of a given radius
-export const circle = (radius=1, start=0) => {
+export const circle = (radius, start=0, x=0, y=0) => {
   let points = []
 
-  for (let i=start; i<=128+start; i++) {
-    let angle = Math.PI * 2.0 / 128.0 * i
-    points.push(new Victor(Math.cos(angle)*radius, Math.sin(angle)*radius))
+  for (let i=0; i<=128; i++) {
+    let angle = Math.PI * 2.0 / 128.0 * i + start
```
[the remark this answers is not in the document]
jeffeb3: It's ok. The new circle packer is the only place where I see anything other than size used.

src/common/geometry.js
```diff
@@ -128,17 +128,36 @@ export const getSliderBounds = (vertices, sliderValue) => {
...
+    points.push(new Victor(x + Math.cos(angle)*radius, y + Math.sin(angle)*radius))
   }
 
   return points
 }
 
+export const arc = (radius, startAngle, endAngle, x=0, y=0) => {
+  let resolution = (Math.PI*2.0) / 128.0 // 128 segments per circle. Enough?
```
jeffeb3, line 143: If I had infinite time, I might change this to be at least some minimum number, but determine it based on the radius. So a small circle might have 16 sides, but a large one might have 256.

src/common/geometry.js
```diff
@@ -128,17 +128,36 @@ export const getSliderBounds = (vertices, sliderValue) => {
...
+  if (deltaAngle > Math.PI) {
+    deltaAngle -= 2.0 * Math.PI
+  }
+  if (deltaAngle < 0.0) {
+    resolution *= -1.0
+  }
+
+  let tracePoints = []
+  for (let step = 0; step < (deltaAngle/resolution) ; step++) {
+    tracePoints.push(Victor(x + radius * Math.cos(resolution * step + startAngle),
```
jeffeb3, line 155: I wonder if we do enough of these calculations to warrant a lookup table.

src/models/circle_packer/CirclePacker.js
```diff
@@ -0,0 +1,357 @@
...
+    const intersection = this.boundaryIntersection(c)
+    const intersection2 = this.boundaryIntersection(end)
+
+    // draw arc to perimeter
+    const a2 = Math.atan2(intersection.y - c.y, intersection.x - c.x)
+    this.points.push(...arc(c.r, angle, a2, c.x, c.y))
+
+    // connect along perimeter
+    const machine = getMachineInstance([], this.settings)
+    this.points.push(...machine.tracePerimeter(intersection, intersection2))
```
jeffeb3, line 324: I found a case where it is tracing across the pattern. It looks like this is where that would be coming from, but I'm not sure what is causing the issue.

screenshot-2021-09-01-1630513422

This is the default 0-500x0-500 rectangle machine.

jeffeb3: Maybe we could make these perimeter points a lot larger? I'm not sure why it is trying to go around to that other corner before tracking on the circle to the left.

jeffeb3: It's not this line of code. This line is fine. I added some logs and these lines aren't causing the issue.

bobnik: This is fixed by #214. I'm going to merge this PR now.

Issue #211, opened 2021-08-23
New lines in text need to be eacaped in the comments


jeffeb3:
A user reported to me that when using multiple lines (with newlines), the new lines aren't being commented out in the comments section, which makes .thr files unusable.

bobnik:
We no longer export layer settings as comments in file exports (save as .sdf file instead), so I'm closing this.

PR #217, opened 2021-12-18
Fancy text

bobnik:
"Fancy" text, using OpenType fonts that are converted into vertices. I've preloaded 9 fonts (all free to use and distribute), but we could add/remove any as desired. Some additional features:

- can choose to connect lines of text along "inside" or "outside"
- alignment

Includes fix for #211.

image


jeffeb3:
This is great.

jeffeb3:
Are all these ttf fonts free and clear for our use? Do we need to attribute them somewhere?

jeffeb3:
The "outside" connection doesn't seem great to me. I can't seem to get it to do anything but swipe right through the words.

I think we should make the inside the default. WDYT?

jeffeb3:
Have you tried breaking up the words into characters to enforce the letter order? It seems to really like going right to left. And Garamond "Jeff" does the J before the f,f,e.

screenshot-2022-01-21-1642790020

jeffeb3:
I found one bug while playing around, but it wasn't related to the fancy text. I will push it here anyway, since it is a tiny bug. I had a long path, I changed the preview slider to the right, and then I added a shorter layer and deleted the longer one, and the page crashes.

jeffeb3:
As long as the fonts are clear, and you don't want to play with the letters, you can merge this.

Is there anything else you'd like to see in a release?

bobnik:
> Are all these ttf fonts free and clear for our use? Do we need to attribute them somewhere?

Free and clear. I will add attribution.

bobnik:
> The "outside" connection doesn't seem great to me. I can't seem to get it to do anything but swipe right through the words.
> 
> I think we should make the inside the default. WDYT?

Inside is probably a better default, but outside works fine for me. I cannot reproduce what you are seeing. I can't reproduce your ordering issues, either. It _does_ break it up letter by letter, and always goes left to right. Can you check and see if you've accidentally set fine-tuning to rotate your starting point - that can cause what you are seeing.

bobnik:
> As long as the fonts are clear, and you don't want to play with the letters, you can merge this.
> 
> Is there anything else you'd like to see in a release?

@jeffeb3 There are some "pesky" bugs introduced with the previous caching refactor. I see occasional crashing, and in addition, layers are not always rendered faithfully. It's inconsistent. I want/need to dig into this prior to a new release. I don't have anything else currently active, so we could consider a release after that?

bobnik:
> > As long as the fonts are clear, and you don't want to play with the letters, you can merge this.
> > Is there anything else you'd like to see in a release?
> 
> @jeffeb3 There are some "pesky" bugs introduced with the previous caching refactor.

@jeffeb3 Ok, I pushed fixes for a few things: a faulty selector was causing issues with redraws when effect layers were added. Also, shapes were being drawn every time a font was loaded (9x), which results in a huge slow down. To fix this, I had to re-structure the entire redux store. It's not actually that big of a change, but it required a lot of refactoring. I actually like the new selectors better - they protect us against future changes a bit more. Anyway, this means that I would ask you to do more than usual amount of testing before approving this PR. Thanks.

jeffeb3:
screenshot-2022-04-30-1651330764

I'm still seeing it draw the letters right to left. To reproduce:
 - I deleted all the local storage
 - I refreshed the page
 - I changed the shape type to fancy text. I left the defaults (see the screenshot).
 - I changed the preview slider to show it was drawing from right to left.

jeffeb3:
The letter order is also causing the "outside" connecting rows to be wrong. If it had ended on the right and started on the left, it would be fine.

screenshot-2022-04-30-1651330976

This is the same test setup, but I added a second row of "Sandify" and then changed it to outside.

jeffeb3:
I opened it in chromium, and it is drawing left to right. How is that possible?

jeffeb3:
I have ran out of time this morning. But I narrowed it down to this:

```
Object.keys(graph)
```
In chromium, it is an array of length 7, containing `['0', '1', '3', '4', '6', '8', '9']`. In firefox, it is `[ "9", "8", "6", "4", "3", "1", "0" ]`. So it must just be storing the keys in a different order. I guess we should sort that array before doing the sort?

jeffeb3:
So, sorting it again makes no sense... The sort should put them in the right order no matter what order they come in as (although I am still not sure why they are in reverse order on firefox). But looking at the doc for the sort function on Array: https://www.w3schools.com/jsref/jsref_sort.asp

It takes two arguments, and returns the difference. So I have a fix that makes sense to me, and works in my two browsers. I will push it in a minute But here is the sort function I am using:

```
    return Object.keys(graph).sort(
      (leftIndex, rightIndex) => {
        const leftPoints = graph[leftIndex].points
        const rightPoints = graph[rightIndex].points
        return Math.min(...leftPoints.map(pt => pt.x)) - Math.min(...rightPoints.map(pt => pt.x))
      })
    .map(key => graph[key])
```

PR #229, opened 2022-05-10
Fix lint

bobnik:
This PR adds a Github action to run lint on push/pull requests, per @geekuillaume's recommendation in #225, and fixes the lint configuration.

@jeffeb3 I tested this locally using act (https://github.com/nektos/act), but unless you have this repo setting enabled, I suspect it won't run and you'll have to test yourself:

image

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -63,7 +63,8 @@ module.exports = {
     '@typescript-eslint/no-useless-constructor': 'warn',
     'react/no-unused-prop-types': 'off',
     'react-redux/no-unused-prop-types': 'warn',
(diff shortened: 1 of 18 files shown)


jeffeb3:
act is really neat. I don't know how I am going to use it, but I will find a way to play with it.

It ran your test code. IIRC, the setting is something like "Don't allow new fork PRs to run". I click a box when someone new makes a PR from a fork, and then they get added to a whitelist. I don't know if it is related, but there was a wave of bots adding workflows to do cryptomining on a bunch of projects in the actions. Which is why I think those settings exist.

PR #248, opened 2022-11-06
Feature/fix gcode import

### LINKED ISSUE #246: gcode import broken
https://forum.v1engineering.com/t/sandify-update-v0-2-7/34684/7?u=jeffeb3

> I think there might be an issue now with importing GCODE files in this new version of [Sandify](https://sandify.org/)? I can import THR, but not GCODE or NC?

We need to look into this. I will try to check it out soon.

Specifically applies to 0.27

bobnik:
This PR fixes #246 by replacing stream-browserify with the needed polyfills.

Please note that Vite was also upgraded in this release, and the default ports changed (5173 for development, 4173 for production preview). I changed the default back in the config to 3000 for compatibility.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,7 +6,7 @@
     <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/css/bootstrap.min.css">
     <link rel="shortcut icon" href="/favicon.ico">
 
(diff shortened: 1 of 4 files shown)


jeffeb3:
This is working great. I tested with the regular run and with the build/preview test locally.

PR #254, opened 2022-11-22
Fine Tuning and Track Transform

jeffeb3:
The track transform isn't exactly the same, but it might just be clearer. My big complaint is that it is a consistent spread in angle, and as the spiral gets bigger, that leads to more stretched out parts. I also didn't reimplement the multiple-shapes-at-each-point, which I thought was neat, but was harder to do here. 

The Fine tuning was easy. I didn't move the connecting method (because I don't have both layers, so I can't). The connection method should probably just move to the Layer class, with the offsets and rotation.

screenshot-2022-11-22-1669135624

screenshot-2022-11-22-1669136982


jeffeb3:
Oops. Wrong from/to branch.

PR #259, opened 2022-12-18
redefine base model classes; restructure directories

bobnik:
@jeffeb3 This is my attempt to reconcile our notes and discussions, and may not represent 100% what we have in our design notes.

- Model (knows about rotation, offsets, other meta-data common to all models)
  - Layer < Model
  - Shape < Layer
  - Group < Layer
  - Effect < Model
- Started to move remaining computer logic into Model. Wasn't much left.
- Fixed issue with group rendering if you changed rotation and offsets.
- Rebuilt the directory structure (mostly) per our notes.
- Added the ability to specify src-local paths starting with '@'

Please take a look and we can chat as needed. I don't want to take this further without consensus.


jeffeb3:
I will carve out some time today to look at this in depth.

jeffeb3:
@bobnik I looked at this. It is in the right direction. I think it follows pretty close to our design discussions. There are still a few things that don't look right to me, but I am not sure if they are things to do in the future, or things that we still need to agree on.

For example, the Model has things like offsetX and offsetY. The layer and effect inherit from Model. I don't think either should. But these both need offsetX and offsetY options. 

So there is still some work to be done (either to continue the changes, or teach me the right way to do it).

I don't see any reason to stop here though. This can be merged, and we can base future work on this.

Issue #276, opened 2023-10-14
Preview render missing when width is 992px


aaronse:
Neat project, nice update!   Noticed preview render is not displayed when width is 992px.  Consider checking out responsive layout breakpoint logic.  Cheers!

bobnik:
Yes, it's all built around responsive breakpoint logic. Seems we're off by a pixel point though lol. Only 992px exactly causes this issue - good catch. We'll fix.

aaronse:
Only noticed because I like how your project is looking as was playing around to see what your responsive layouts look like :-) 
Thoughts on https://github.com/jeffeb3/sandify/pull/277 ?

aaronse:
Built and verified fix... 

![image](https://github.com/jeffeb3/sandify/assets/16479976/f6fabb38-0c8e-457b-bbf0-deb600bffc50)

bobnik:
Fixed by #277

PR #277, opened 2023-10-14
minor fix to breakpoint logic https://github.com/jeffeb3/sandify/issues/276

aaronse:
minor fix to breakpoint logic https://github.com/jeffeb3/sandify/issues/276

### DIFF
--- a/src/features/app/App.scss
+++ b/src/features/app/App.scss
@@ -39,7 +39,7 @@
   width: 550px;
   padding-top: 0.5rem;
 
-  @media not (min-width: 992px) {
+  @media (max-width: 991px) {
     width: 100%;
   }
 }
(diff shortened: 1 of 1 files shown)


src/features/app/App.scss
```diff
@@ -39,7 +39,7 @@
   width: 550px;
   padding-top: 0.5rem;
 
-  @media(max-width: 992px) {
+  @media not (min-width: 992px) {
```
bobnik, line 42: Instead of using "not" here, just change min-width to 991px

aaronse: not seemed like more robust way to avoid this issue in the first place?

bobnik: Disagree. Breakpoint is exactly at 992px. No need for a more general case, and I do not find "not min-width" to be intuitive. You're welcome to make the change, or I can fix independently.

aaronse (commented):
Applied feedback from @bobnik

PR #279, opened 2023-10-15
ensure that saved file imports can be reloaded

bobnik:
Bug: Cannot reload a saved file import layer. It throws an exception. To reproduce, import an exported pattern as a layer and reload the page. The pattern gets reset. The same thing happens if you save to sdf and try to load the file.

This PR fixes the issue.


jeffeb3:
This has been incorporated at some point. Closing.

PR #289, opened 2024-05-11
update packages

bobnik:
@jeffeb3 This PR updates all possible packages. A few couldn't be updated due to dependencies. It builds and runs fine for me, but I did not do extensive testing.


jeffeb3:
Thanks for doing that. I did some testing and it looks good. The changes make sense.

PR #291, opened 2024-06-07
Feature/image import

bobnik:
This is an alternative interpretation of the idea presented in #174, importing images as plotter lines. I used https://mitxela.com/plotterfun/ as the basis for the plotting. Images are loaded and stored as data URLs within the state (and are also exported in SDF files).

To import an image, select 'File, Import image...".

As you'll see, imported images require some fiddling to get looking okay. I'm going to test a few samples on my table to see if the defaults need adjusting, but please feel free to suggest values.


jeffeb3:
This is pretty fun and definitely a value add. I can't seem to get the sawtooth to work on the images I've tried. It looks like it maybe has a math error in it. I was mostly playing with high contrast images like this batman logo: https://i.pinimg.com/originals/1e/43/01/1e4301c5b23b4e54cf6b79743f8bf6a2.png

I tried an svg, but that wasn't supported. It seems like logos or cartoons would be ideal for this format. 

The plotterfun seems to work much better and is much quicker. Why is that site so fast and ours is so slow? Would it make sense to push people into that site and then support importing svg line art instead? I'd like to fine tune the amplitude or contrast, but it takes so long to render, it just isn't worth it. We should probably add the delayed execution (like we do with the math operations) to make it not get stuck updating 5 times when you change the value by 0.1 increments 5 times.

I tried some photos with our family's faces, but that wasn't that interesting. Maybe it would look better if I plotted it. I would have to know which photo it was for me to be able to know who it was. I can imagine knowing it was a dog, but not which dog.

The FM squiggles sometimes had funky spacing on pure white areas. But that may just be part of the math. It probably switches to zero frequency change and keeps whatever phase it was set to when it stopped. The AM only seems better in the preview, but I don't know how that will translate to the sand. Did you have any luck with that?

bobnik:
Hmm. It runs quickly for me, but I am on a very beefy Apple Silicon
Macbook. Plotterfun uses a webworker to calculate things asynchronously,
which I didn't.

I'll take a look at your various points this weekend.

P.S. It's generally considered a bad idea to rebase a remotely pushed PR. I
can find an article explaining if needed. You can do a merge into the PR
instead and push that up before merging into the main branch.

jeffeb3:
th-1201244194

Here is another picture type (landscape). From another image search for maroon bells.

I'm not expecting miracles with these. I am just trying to build a small suite of examples so we know what to expect (and I don't want to paste my family's face everywhere).

I am leaning towards building defaults that are better for logos than for images. The images just aren't that compelling to me, and I expect them to take a lot of tweaking to look right. But I expect a batman logo, or a brewery/bar logo to look good enough to be recognizable. Ideally without a lot of tweaking.

bobnik:
Re-reading your emails, I really don’t see how it’s so slow for you. Changing settings for me is instantaneous. I suspect I must be using much smaller images. One way I think to speed things up is to downscale the images. I will play around this weekend.On Jun 13, 2024, at 3:07 PM, Jeff Eberl ***@***.***> wrote:﻿
th-1201244194.jpeg (view on web)
Here is another picture type (landscape). From another image search for maroon bells.
I'm not expecting miracles with these. I am just trying to build a small suite of examples so we know what to expect (and I don't want to paste my family's face everywhere).
I am leaning towards building defaults that are better for logos than for images. The images just aren't that compelling to me, and I expect them to take a lot of tweaking to look right. But I expect a batman logo, or a brewery/bar logo to look good enough to be recognizable. Ideally without a lot of tweaking.

bobnik:
@jeffeb3 As mentioned, I think that the performance and rendering issues are fixed.

After drawing various images on the table, I've realized (as you did) that most images do not have enough contrast to produce interesting sand designs. The height of the ball as it moves through the sand on each line is too relatively large, and subtle "squiggles" are relatively small and get lost in the line and make a muddled mess.

Ideally, we would be able to infer what parts of the image are background noise and can be "clipped", allowing us to focus on patterns within the image. I have some vague ideas on how to approach this, so if it's ok with you, let's put this PR on hold for now and I'll circle back after some experimentation.

bobnik:
@jeffeb3 I'm done with my additional changes and think this is ready to merge, barring any issues you find. I've added a slider component, used by the image import for now, but I can see us using it in other shapes. In addition, there is a new "Brightness filter" option for some of the plotter types that filters out background. It works best when the background is solid. I think I could eventually get it working for all types, but I hit some snags and decided not to delay this PR any further.

image

jeffeb3:
I had some time to check this out today. It is looking really good. It's much faster for me, and the example images seem to do reasonable things. This is great.

PR #292, opened 2024-06-08
Add Noto Emoji font

bhamiltoncx:
My 10-year-old kiddo has been begging me to add some emoji-based designs to our new Sisyphus Xyla.

This PR imports [the Noto Emoji font](https://fonts.google.com/noto/specimen/Noto+Emoji) and adds it to the Fancy Text font drop-down.

Example result:

Screenshot 2024-06-08 at 1 09 06 PM


bobnik:
Ha! This looks great and will be a welcome addition.

bobnik:
This will be more friendly to use via phone. Probably just me, but I had to look up how to add an emoji via keyboard on a Mac (Control + Command + space). I guess it can be done via browser menu, too.

bhamiltoncx:
Thanks for the speedy review and merge!

> This will be more friendly to use via phone. Probably just me, but I had to look up how to add an emoji via keyboard on a Mac (Control + Command + space). I guess it can be done via browser menu, too.

Yeah, I usually enable the Input menu in the menu bar in Settings:

Screenshot 2024-06-08 at 1 45 20 PM

That lets me show the emoji picker directly using the mouse:

Screenshot 2024-06-08 at 1 44 50 PM

bhamiltoncx:
By the way, I fruitlessly spent probably 4-5 hours trying to get [the Noto Color Emoji font](https://fonts.google.com/noto/specimen/Noto+Color+Emoji) to work. Learned a ton about OpenType but it was a waste of time in the end. :)

In case you or anyone else goes down the same wrong path as me, note that [`opentype.js` only recently started supporting](https://github.com/opentypejs/opentype.js/pull/695) [the OpenType COLR table](https://learn.microsoft.com/en-us/typography/opentype/spec/colr) used by color emoji fonts.. but it still currently only supports [the older COLR version 0 standard](https://learn.microsoft.com/en-us/typography/opentype/spec/colr#colr-header) and not the COLR version 1 standard used by Noto Color Emoji. :/

Thankfully, Noto Emoji is a regular old font and doesn't need any special logic in `opentype.js`.

PR #294, opened 2024-06-13
Feature/shape descriptions

bobnik:
This PR adds longer shape descriptions so that those who are interested can learn more about the algorithms and equations we use to build our patterns. It also moves some licensing info around for consistency. Lastly, it adds some carriage returns for better formatting in a few files.


jeffeb3:
This looks good. I added some minor text changes and rebased it (and I deleted it for a minute, which broke this PR).

Issue #155, opened 2020-05-15
Randomize Equation Button


jeffeb3:
Similar to @msurguy's flow lines randomize button: https://msurguy.github.io/flow-lines/, it would be neat to have some random equations for scale and spin to make a lower barrier of entry to editing the functions.

Original comment from Maksim: 
> Would be cool to have Randomize button for scale and spin. I got my randomizer from @anvaka's repository here: https://github.com/anvaka/fieldplay/blob/master/src/lib/generate-equation.js , here's my take on it, with math.js functions: https://github.com/msurguy/flow-lines/blob/master/src/lib/function-generator.js

PR #301, opened 2024-10-13
Feature/randomize values

bobnik:
This PR was inspired by #155, but it doesn't actually address the feature proposed there (though it could be easily added).

I've added a new "randomize value" button (die) for layers and effects. It makes sensible randomized choices for all layer and effect settings. It's pretty fun to play around with. 

Random defaults and ranges can obviously be adjusted as needed as we play around with it.

### DIFF
--- a/src/common/Model.js
+++ b/src/common/Model.js
@@ -1,3 +1,5 @@
+import { functionValue } from "@/common/util"
+
 const options = []
 
 export default class Model {
@@ -10,6 +12,7 @@ export default class Model {
       usesFonts: false,
       dragging: false,
       stretch: false,
+      randomizable: true,
     })
   }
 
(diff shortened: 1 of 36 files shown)


jeffeb3:
I didn't test all of these changes, but the randomize is pretty fun. Especially with shapes like the fractal spirograph where most options are interesting and there is a significant dynamic to changing the options.

I was going to suggest putting it on the shape options itself, instead of the new modal. But I worry that someone will accidentally clobber a bunch of work if they click it and don't have an undo button.

bobnik:
> I was going to suggest putting it on the shape options itself, instead of the new modal. But I worry that someone will accidentally clobber a bunch of work if they click it and don't have an undo button.

@jeffeb3 It _is_ also in the shape options, though it's a little "dice" icon. You can keep clicking it to get different results each time.

Issue #256, opened 2022-11-26
Starting GCode getting reformatted


V1EngineeringInc:
When I enter this starting GCode it always gets reformatted, in a way that the machine will not run properly. I have to manually edit all my files before running them.

```
; BEGIN PRE

$HY

G0 Y-4 F50

G92 Y0

$HX

G0 X-13 F50

G92 X0

G1 X1 F1000

; END PRE
```

That is what I need,

```
; BEGIN PRE

$HY G0 Y-4 F50

G92 Y0

$HX

G0 X-13 F50

G92 X0

G1 X1 F1000

; END PRE
```

That is what I get. I have to edit that first line. Any ideas? I made sure I did not have any extra spaces or anything.  The second $HX command seems to work fine so not a "$" thing?

V1EngineeringInc:
Had to add line breaks to get it to look okay. It comes out with no spaces, as expected from Sandify.

jeffeb3:
I was worried it was the `$HY`, because I didn't think those were allowed in gcode. But it isn't that, it is the first line always getting joined with the second:

```
One
Two
Three
```
Becomes:
```
; BEGIN PRE
One Two
Three
; END PRE
```

Thanks for reporting it. It should be an easy fix.

V1EngineeringInc:
Well, bummer it is a bug, but glad I wasn't just doing something wrong.

I have a fun gcode running right now, can't wait to show you....and it lead to an idea.

jeffeb3:
If you just add an empty line as the first line, it will be fine. 

Bob and I are working on a bunch of changes at once in a restructuring to enable some new features. IDK when that will be finished. I'm interested to see what you've come up with, and make sure it fits in the new system.

V1EngineeringInc:
Nice tip I will do that.

bobnik:
Fixed in #302

Issue #237, opened 2022-08-09
Add new "code" layer


michaelcunningham1974:
It would be useful to have a new layer type called code that would provide a textbox that arbitrary code could be entered similar to the start/end code in the machine settings.

I envision this being useful to do things like pauses  between layers, changing federates, changing led settings, custom pin/io etc.

jeffeb3:
This is several months old now, I'm not sure how I missed it. 

That would be easy enough to add. My concern is just being able to get the user to understand it. We wouldn't be able to have any kind of preview for that text, it would just be a layer that gets copy/pasted at the end. I'm not 100% sure our architecture can manage that kind of storage through to the download ATM either.

bobnik:
Added in #302

Issue #236, opened 2022-08-09
Add Comment Line Between Tracks on Export


michaelcunningham1974:
Currently when exporting multiple layers, Sandify just continues appending gcode to the end of the file. I suggest adding a delimiter comment at the start/end of each layer something like:

;Layer Start: [LAYERNAME]

...GCODE...

;;Layer Start: [LAYERNAME]

That way if we want to run a post processor on the code we can isolate each layer.

bobnik:
Added in #302

Issue #216, opened 2021-09-23
Layer changes in the pattern files


mrehorst:
After trying to format some text on my table using multiple layers I've noticed that there's no indication within the pattern files of where one layer ends and the next begins. It could be useful, especially for tweaking the line position that starts text, if there was some comment placed in the pattern file to indicate when a layer change occurs.  That would allow some manual editing or even automated post-processing to do additional things at the layer changes.

jeffeb3:
I like that idea. I will take a look and see if that is easy in the code. I suspect we are calling one big "calculate" function with the whole state and getting back a bunch of vertices. But maybe it will be easier than I think.

bobnik:
Added in #302

PR #302, opened 2024-10-15
annotated exports

bobnik:
This PR addresses the following issues:

#256 Starting GCode getting reformatted
#237 Add new "code" layer
#236 Add Comment Line Between Tracks on Export
#216 Layer changes in the pattern files

New features:

1. The start and end of each layer and connector is noted in exports as comments.
2. New "program code" effect which adds ability to insert PRE/POST code logic around layers. Variable substitution is also supported in these blocks, local to the given layer (e.g., maxx, maxy).

To test the export format changes, export a pattern in different formats (and ideally compare with older exports).
To test the new effect, add one to a shape and add some sample GCODE or instructions before you export.

This PR introduces a likely bug (or at least a difference) in the SCARA output. If you do a before/after PR export, you'll see the x/y value differences. This occurs because I've added logic to normalize theta during export between 0 and 2PI. It works fine with .thr files, as the exported theta values are equivalent. @jeffeb3 can you let me know if the changes here to the SCARA outputs are broken? We can certainly rollback the normalization, but I think it's generally a good idea.

### DIFF
--- a/src/common/geometry.js
+++ b/src/common/geometry.js
@@ -365,13 +365,6 @@ export const downsample = (vertices, tolerance = 0.0001) => {
   return result
 }
 
-// Normalize the theta value to the range [0, 2*pi)
-export const normalizeTheta = (theta) => {
-  const TWO_PI = Math.PI * 2
-
-  return ((theta % TWO_PI) + TWO_PI) % TWO_PI
-}
-
 // Convert x,y vertices to theta, rho vertices
 export const toThetaRho = (subsampledVertices, maxRadius, rhoMax) => {
   let vertices = []
@@ -396,7 +389,9 @@ export const toThetaRho = (subsampledVertices, maxRadius, rhoMax) => {
       subsampledVertices[next].x,
       subsampledVertices[next].y,
     )
-    rawTheta = normalizeTheta(rawTheta)
+
+    // Convert to [0, 2pi]
+    rawTheta = (rawTheta + 2.0 * Math.PI) % (2.0 * Math.PI)
 
     // Compute the difference to the last point.
     let deltaTheta = rawTheta - previousRawTheta
@@ -413,7 +408,7 @@ export const toThetaRho = (subsampledVertices, maxRadius, rhoMax) => {
 
     previousRawTheta = rawTheta
     previousTheta = theta
-    vertices.push(new Victor(normalizeTheta(theta), rho))
+    vertices.push(new Victor(theta, rho))
   }
 
   return vertices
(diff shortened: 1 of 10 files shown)


jeffeb3:
> It works fine with .thr files, as the exported theta values are equivalent. 

That is not true. This is an important part of the thr output. If one vertex is -0.1pi and the next is 1.9pi, it will not shortcut through zero, it will go around the circle. So normalizing either the scara or thr would change the output.

The example Bruce uses as an example for the simplest thr wiper is:
```
0 0
628 1
```
which goes 100 loops around and goes smoothly from 0-1 radius.

You can normalize the theta before computing deltaTheta, but after deltaTheta, you have to keep the previous + deltaTheta without normalizing it again.

Why did you feel the need to normalize between those angles?

src/common/geometry.js
```diff
@@ -407,7 +413,7 @@ export const toThetaRho = (subsampledVertices, maxRadius, rhoMax) => {
 
     previousRawTheta = rawTheta
     previousTheta = theta
-    vertices.push(new Victor(theta, rho))
+    vertices.push(new Victor(normalizeTheta(theta), rho))
```
jeffeb3, line 416: I think this change just needs to be reverted.

jeffeb3:
Maybe the motivation was to have min/max theta make more sense. I would leave those as-is. Because I can imagine wanting to use max or min theta, or really start or end theta, in some special finishing move. But if you want the normalized version, maybe make new parameters for minNormTheta and maxNormTheta and only normalize them before you add them to the stats.

src/common/geometry.js
```diff
@@ -365,6 +365,13 @@ export const downsample = (vertices, tolerance = 0.0001) => {
   return result
 }
 
+// Normalize the theta value to the range [0, 2*pi)
+export const normalizeTheta = (theta) => {
+  const TWO_PI = Math.PI * 2
+
+  return ((theta % TWO_PI) + TWO_PI) % TWO_PI
```
jeffeb3, line 372: I'm guessing this slightly different form has to do with an edge case near zero, but slightly negative?

bobnik: @jeffeb3 Thanks for the clarification on this. I'll revert all normalization logic.

bobnik:
@jeffeb3 I've pushed a commit to preserve theta across layers. I'm seeing identical numbers in the before/after exports, so I'm hopeful this addresses your question.

jeffeb3:
Cool. I will check it out tomorrow or Thursday.

jeffeb3:
I don't have a good way to test this. I don't have a SCARA machine. But since the values are the same, I can't see how it could be broken. Thanks.

Issue #285, opened 2024-01-10
rho in thr format is limited to 0.5 with rectangle table 


vincentbraillard:
When exporting in .thr format with a square/rectangle table, the rho is limited to 0.5. why?
it could be better if 1 = corner of the rectangle....

bobnik:
Fixed in #303

PR #303, opened 2024-10-16
fix maxRadius calculation

### LINKED ISSUE #222: Incorrect scale with exporting THR for rectangular table
I have a couple of requests and a potential issue.

Let me start with the issue. I have built a CoreXY [Sand-Table](https://github.com/rdudhagra/Sand-Table) based on the design by [rdudhagra](https://github.com/rdudhagra), and the design calls for .thr files. In Sandify, if I set the machine to Rectangular machine, Max X (mm) 450 and Max Y (mm) to 450, my table, draw a design that fills the table and export it as .thr, when I play the track on my table it is nowhere near the correct size. To double-check, when I import the .thr file to Sandify, it becomes smaller than what I designed but is about the size my table appears to draw.

However, if I use the settings for a Rectangular machine and export it as Gcode, then import the file for review, it appears as the correct size and doesn't shrink. The same also applies for a Polar table; if I set the machine to Polar machine, export as a .thr file, then import, it appears to be the correct size and doesn't shrink. It just doesn't allow for .thr to be exported to the correct size for a Rectangular machine. Is it possible to export .thr files for a Rectangular machine that has the correct dimensions? Did I miss something? If not, can it be possible to export and import .thr designs for a Rectangular machine? 

My second request, could the same features for the Start and End points of a Polar table also be added to a Rectangular table?

Thanks!

bobnik:
Fixes #222 and #285.

I believe this is as easy as correctly calculating the maxRadius when exporting to a rectangular table. We were calculating a radius double the actual maximum. This bumps the max rho to 1 in the exported file.

### DIFF
--- a/src/features/export/ExportDownloader.js
+++ b/src/features/export/ExportDownloader.js
@@ -53,8 +53,8 @@ const ExportDownloader = ({ showModal, toggleModal }) => {
     maxRadius:
       machine.type === "rectangular"
         ? Math.sqrt(
-            Math.pow((machine.maxX - machine.minX)/2, 2.0) +
-              Math.pow((machine.maxY - machine.minY)/2, 2.0),
+            Math.pow((machine.maxX - machine.minX) / 2, 2.0) +
+              Math.pow((machine.maxY - machine.minY) / 2, 2.0),
           )
         : machine.maxRadius,
     vertices: useSelector(selectConnectedVertices),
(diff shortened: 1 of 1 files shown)


vincentbraillard:
Hi. Yes the correction you did should works. Have you test it. I don't no how to test other branch than the one on the web ;-(

bobnik:
> Hi. Yes the correction you did should works. Have you test it. I don't no how to test other branch than the one on the web ;-(

No worries! We'll get it tested and merged soon.

PR #306, opened 2024-11-03
Voronoi shape

bobnik:
This PR adds a Voronoi shape with a number of different parameters that produce pleasing results. 

image
image
image

There is also a Voronoi effect, which is spectacular (imo), but it's too slow to include by default as it can hang the browser. If you get a chance, can uncomment the lines to test it out. We can add it if/when we figure out how to move vertex computation into a worker thread. I've been looking into that, but I didn't want to hold up this PR in the meantime.


jeffeb3:
I want to do more tests and playing with this. But this was opened in November and I'm just looking at it now. That's not fair, so I am merging it based just on inspection. It looks cool. I will need to test it before we make a new release (and this will be a gem, for sure).

As for the worker thread. That sounds like a good idea. I would also like to try web assembly for things like the effects (one at a time, probably).

PR #312, opened 2025-01-25
add id to cache key

### LINKED ISSUE #309: Cache issue on import new Layer
I found an issue where the import seems to be caching previous file imports if the name is the same. 
I was using Sandify to verify THR output from another program that spits out "output.thr". Tested file #1, didn't love the result so I went back and ran it again with different parameters. The program created a new file called "output.thr". When I input that into Sandify, it still used the data from the first input. If I rename the second file "output2.thr", it imports just fine.

bobnik:
This fixes #309


jeffeb3:
This looks like a good fix. Thanks.

PR #317, opened 2025-07-14
fix js error due to missing variable

bobnik:
SVG export is currently broken, and this PR fixes the issue.


jeffeb3:
Ah shoot. I didn't test that, I'm sure. I wonder how many people have been bitten by it.

PR #319, opened 2025-07-20
Add Umami hack back in.

jeffeb3:


jeffeb3:
@bobnik, This should fix umami.is. I need to add this to the main index.js somewhere, so it gets built in the npm output.

PR #327, opened 2026-01-01
voronoi effect

bobnik:
Adds a "uniformity" slider to Voronoi patterns that applies Lloyd relaxation, making cells more evenly distributed. Also improves path-finding performance.

**New option:**
- **Uniformity** (0-20) - applies Lloyd relaxation iterations to make Voronoi cells more regular/uniform
- Available on both Voronoi shape and Voronoi effect

**Performance improvements:**
- Binary heap PriorityQueue - O(log n) enqueue/dequeue (was O(n))
- Dijkstra restored with proper PriorityQueue - O((V+E) log V)
- BFS for unweighted shortest path - O(V+E)
- KDBush spatial index for densityWeight - O(n log n) (was O(n²))
- Removed redundant `clearCachedPaths()` calls on addNode/addEdge

| Optimization                  | Before         | After          | Speedup             |
  |-------------------------------|----------------|----------------|---------------------|
  | PriorityQueue enqueue/dequeue | O(n)           | O(log n)       | ~10x for 100 items  |
  | Dijkstra (overall)            | O(V² + VE)     | O((V+E) log V) | ~5-10x typical      |
  | densityWeight                 | O(n²)          | O(n log n)     | ~37x for 300 points |
  | BFS for unweighted            | O((V+E) log V) | O(V+E)         | ~2-3x               |

### Test plan
- [ ] Verify Voronoi shape renders with uniformity = 0 (unchanged behavior)
- [ ] Test uniformity slider at various values (5, 10, 20)
- [ ] Verify Voronoi effect also has uniformity option
- [ ] Check performance on complex patterns (BFS should be faster)

### DIFF
--- a/src/common/lindenmayer.js
+++ b/src/common/lindenmayer.js
@@ -17,7 +17,7 @@ const shortestPath = (nodes) => {
     if (visited[edge12Key]) {
       const result = nearestUnvisitedNode(i + 1, nodes, visited, graph)
 
-      if (unvisitedNode != null) {
+      if (result != null) {
         const shortestSubPath = graph.bfsShortestPath(
           node1Key,
           result.node.toString(),


jeffeb3:
Also awesome. This looks much more like scales on a reptile or something. People are going to enjoy adjusting that uniformity.

bobnik:
> Bug 1 found: If I change the combobox (the input box on the right), it doesn't change anything. I'm using firefox and the combobox has little arrows for going up or down, and the scroll mouse wheel also affects that number. When I change that, the slider changes position, but the pattern doesn't change unless I actually click the slider.

There's a fix for the slider in #326

bobnik:
> Bug 2: Default is zero sized. I'm not sure if this is a new thing. But if the last shape I added was a voronoi, and I refresh the page, then select file...new, it defaults to a voronoi shape with width and height set to 0,0.
> 
> I noticed this because I blew away the localstorage and switched branches. But those steps seem to reproduce it without messing with localstorage or git/npm.

@jeffeb3 This is because the machine's not available when the shape is drawn after the reset. I could add code to make this work, _or_ set a sensible default like 100 x 100, but that would be inconsistent with the machine sizing you get when you add a new Layer. Lastly, we could avoid fixing this by waiting until we add your desired feature to clear all shapes with "File, New", which will make the issue go away. If you like that idea, this is ready to merge.

PR #329, opened 2026-01-01
Future ideas

bobnik:
Adds an IDEAS file to enumerate lots of exciting future shapes and effects for us to build. Can extend this file in the future if we want to include other enhancements.

View markdown [here](https://github.com/jeffeb3/sandify/blob/06df1d87968deb8aea9af5f0faf6d9e3c2829daf/IDEAS.md).


IDEAS.md
```diff
@@ -0,0 +1,671 @@
...
+
+- [Shape Ideas](#shape-ideas)
+- [Effect Ideas](#effect-ideas)
+- [Rejected Shape Ideas](#rejected-shape-ideas)
+
+---
+
+## Shape Ideas
+
+Candidates for new shape types. All are algorithmic, render in 2D, and maintain integrity when optimized for contiguous drawing on a sand table. In addition, this provides a new list of categories that can be used to classify shapes in the UI during selection (more precise than our current **Shape** and **Eraser** categories).
```
jeffeb3: > In addition, this provides a new list of categories that can be used to classify shapes in the UI during selection
I don't really like these category names. I agree that "eraser" has lost it's edge. Maybe "Fill Patterns" would be better. Adding an additional category, like "Import" would be good too. But "Constructions", "Fields" sound too dry to me.

bobnik:
I added it in markdown mainly because I had some LLM assistance, which makes it super easy to compile, make broad changes, and (later) keep it up to date. This doesn't need to be the final word on new shapes, obviously, but I wanted a place to capture all the possibilities (and there are _so_ many!) in one place for reference. I'm open to moving it (md format preferred). I definitely agree that we won't be fleshing out details or collaborating in this document.

You'll notice a trend in some of the other PRs, too, where I add a shape or feature-specific README to capture the basics, mainly for contributors who are interested.

PR #330, opened 2026-01-02
Add i18n support with Chinese translations (adapted from @songfei's PR)

bobnik:
This PR adds internationalization support to Sandify, adapted from @songfei's i18n implementation on `songfei/master`. All Chinese translations (~250 strings) are from @songfei's original work.

@jeffeb3, please weigh in on the approach and happy to adjust if needed.
@songfei, special thanks for your contribution. This is an awesome addition to Sandify. Let me know if I've missed anything in the adaptation.

#### What's included

- **i18next integration** with React bindings
- **Chinese translations** for all UI labels, options, tooltips, and shape descriptions
- **Language selector** on the Settings tab
- **Browser language detection** - auto-detects Chinese browsers, respects saved preference
- **Graceful fallback** - missing translations display English

[README.md](https://github.com/jeffeb3/sandify/blob/c7da66af5e3e84440c3b03efdd7668034deb3638/src/i18n/README.md) details how to add future language support.

#### Approach

We use English strings as translation keys (e.g., `t("Save")`). This works well for Sandify since English is the primary development language - keys are readable, greppable, and missing translations gracefully fall back to the English key itself. For longer strings, we define semantic keys (e.g., `t("description.heart")`) and provide translations across all languages.

#### Not included (future PR)

Chinese FancyText fonts: The fonts are quite large, and highlight a larger issue of the application loading all fonts into memory upon load. We'll add a future PR with on-demand loading for fonts which will speed up application load time.

#### Testing

- [ ] Switch language in Settings tab, verify UI updates
- [ ] Refresh page, verify language persists
- [ ] Clear localStorage, verify browser language detection works
- [ ] Navigate between tabs after language change, verify patterns render

### DIFF
--- a/src/features/effects/NewEffect.js
+++ b/src/features/effects/NewEffect.js
@@ -45,7 +45,7 @@ const NewEffect = ({ toggleModal, showModal }) => {
     label: t(opt.label),
   }))
   const [type, setType] = useState(defaultEffect.type)
-  const [name, setName] = useState(defaultEffect.label)
+  const [name, setName] = useState(t(defaultEffect.label))
   const [randomize, setRandomize] = useState(false)
   const selectedEffect = getEffect(type)
   const selectedOption = {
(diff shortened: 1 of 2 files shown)


songfei:
I'm very happy with the changes based on my pull request. However, I noticed that this modification doesn't include the internationalization of the `shapes` and `effects` properties, which I had already implemented in my original code.

bobnik:
@jeffeb3 Added the zero stats fix to #331, so I'm merging this.

